PR #23, opened 2020-07-25
Trackbar

### LINKED ISSUE #9: Trackbar UI on bottom
Add track bar on bottom

May be difficult to sync up with time currently

Goals:
- Be able to move middle tracking thing left and right of trackbar
- Console.Writeline position of the middle tracking thing, 0 is leftmost and 1 is rightmost
- Think about how we can change Story's GameTime from changing the middle tracking thing value

### LINKED ISSUE #20: Seek the game time and music time by clicking on trackbar
- May want to follow osu!'s example where possible, this might be inheriting like from a component and then customizing functionality on top of that
- Maybe it's more lightweight to do something like React-like/JavaScript-like, where you pass a function into the trackbar, and that component runs that function whenever it changes its value
- Or do something like binding the trackbar's value with the GameTime directly somehow, idk, that sounds kind of jank, but some frameworks probably do this

Naranja-Sagged:
Closes #9 
Closes #20


S2VX.Game/Timeline.cs
```diff
@@ -0,0 +1,86 @@
...
+        private const float slider_height = 0.5f;
+        private Container LineContainer { get; set; }
+
+        [BackgroundDependencyLoader]
+        private void load()
+        {
+            RelativeSizeAxes = Axes.Both;
+            InternalChildren = new Drawable[]
+            {
+                LineContainer = new Container
```
maxrchung, line 41: If we don't need to save LineContainer variable, then remove it

S2VX.Game/Timeline.cs
```diff
@@ -0,0 +1,86 @@
...
+                {
+                    Name = "Timeline content",
```
maxrchung, line 43: Check if this really needed, rather not make properties if we aren't going to use them, maybe it's used for debug?

S2VX.Game/Timeline.cs
```diff
@@ -0,0 +1,86 @@
...
+        {
+            Name = "Slider",
+            Colour = Color4.White,
+            Height = slider_height,
+            Width = timeline_width / 150, //600
+        };
+
+        private const float timeline_height = 0.075f;
+        private const float timeline_width = 1.0f;
+        private const float slider_height = 0.5f;
```
maxrchung, line 32: Variable only being used once, think it's better to just put it directly into TimelineSlider

S2VX.Game/TimelineBar.cs
```diff
@@ -0,0 +1,100 @@
...
+
+namespace S2VX.Game
+{
+    public class TimelineBar : Box
+    {
+        [Resolved]
+        private Story Story { get; set; }
+        [Resolved]
+        private TimelineSlider TimelineSlider { get; set; }
+        private float lineWidthUnit = 2f / 3f;
```
maxrchung, line 20: Double f not necessary

S2VX.Game/TimelineBar.cs
```diff
@@ -0,0 +1,100 @@
...
+
+        private double trackLength => Story.Track.Length;
```
maxrchung, line 22: For notation, I think having a comment on why this is a function would be nice

S2VX.Game/TimelineBar.cs
```diff
@@ -0,0 +1,100 @@
...
+using osu.Framework.Graphics.Shapes;
+using osu.Framework.Input.Events;
+using osuTK.Graphics;
+
+namespace S2VX.Game
+{
+    public class TimelineBar : Box
+    {
+        [Resolved]
+        private Story Story { get; set; }
```
maxrchung, line 17: Try and always set declarations = to something

Maybe prefer
`private Story Story { get; set; } = new Story()`

Ideally, it's best practice to never initialize something as null, but sometimes in the case of asynchronous loading this is harder to avoid

I don't think the above will work because Story may reference itself, in that case we can set this to null
`private Story Story { get; set; } = null`

S2VX.Game/TimelineBar.cs
```diff
@@ -0,0 +1,100 @@
...
+        protected override bool OnDragStart(DragStartEvent e) => true;
+
+        protected override void OnDrag(DragEvent e)
+        {
+            if (!delayDrag)
+            {
+                var xPosDelta = (DrawWidth - (DrawWidth / 1.5f)) / 2;
+                var newX = X + ToLocalSpace(e.ScreenSpaceMousePosition).X - xPosDelta;
+                var lineWidth = lineWidthUnit * Story.DrawWidth;
+                var xLengthRatio = newX / lineWidth;
```
maxrchung, line 62: For code reuse, probably best to extract any reused calculations into a function that maybe modifies member variables or whatever you best think of

S2VX.Game/TimelineSlider.cs
```diff
@@ -0,0 +1,22 @@
...
+using osu.Framework.Allocation;
+using osu.Framework.Graphics;
+using osu.Framework.Graphics.Shapes;
+using osu.Framework.Input.Events;
+
+namespace S2VX.Game
+{
+    public class TimelineSlider : Box
+    {
+        private float lineWidthUnit = 2f / 3f;
```
maxrchung, line 13: ??

S2VX.Game/TimelineSlider.cs
```diff
@@ -0,0 +1,22 @@
...
+using System.Collections.Generic;
+using System.Text;
+using osu.Framework.Allocation;
+using osu.Framework.Graphics;
+using osu.Framework.Graphics.Shapes;
+using osu.Framework.Input.Events;
+
+namespace S2VX.Game
+{
+    public class TimelineSlider : Box
```
maxrchung, line 11: Would this be better as RelativeBox since you're using the Relative properties

S2VX.Game/TimelineSlider.cs
```diff
@@ -0,0 +1,22 @@
...
+    {
+        private float lineWidthUnit = 2f / 3f;
+
+        public TimelineSlider()
```
maxrchung, line 15: Per osu! guidelines, may be better to use private load() function

S2VX.Game/TimelineSlider.cs
maxrchung: Per our discussion, this class may not be necessary since it's so small

S2VX.Game/Timeline.cs
maxrchung: We'll try and keep these going forward since it helps nicely with debug

PR #86, opened 2020-08-26
Closes #79 Delete selected note

### LINKED ISSUE #79: Delete selected note
Delete selected note

Xenocidel:
Deletes a note that is selected when the delete key is pressed.


S2VX.Game/Editor/S2VXEditor.cs
```diff
@@ -185,6 +185,11 @@ public class S2VXEditor : CompositeDrawable {
                     break;
                 case Key.Right:
                     NotesTimeline.SnapToTick(false);
+                    break;
+                case Key.Delete:
+                    if (ToolState is SelectToolState tool) {
```
Naranja-Sagged, line 190: I didn't know type checking like this was possible, lol. If there were other keyboard inputs for the other Tool states, I would probably prefer having a virtual OnKeyDown added to ToolState and call ToolState.OnKeyDown at the top, but in this case I think it's fine.

maxrchung: I believe the right way to handle this is to call a ToolState.OnKeyDown at the top. This lets a tool handle the key down internally if it supports it. At the top, you should call ToolState.OnKeyDown, and if this returns true, you don't need to further process key inputs within S2VXEditor.

S2VX.Game/Editor/S2VXEditor.cs
```diff
@@ -185,6 +185,9 @@ public class S2VXEditor : CompositeDrawable {
                     break;
                 case Key.Right:
                     NotesTimeline.SnapToTick(false);
+                    break;
+                case Key.Delete:
+                    ToolState.OnToolKeyDown(e);
```
maxrchung, line 190: As discussed, this should be moved to the top of the handler so that ToolState can handle the key and return back whether a key was processed.

PR #96, opened 2020-08-30
77 note should appear behind grid

### LINKED ISSUE #77: Note should appear behind grid
This needs to be done if possible without separating Note and Approach into 2 separate Drawable components.

May want to take a look at the Depth property of a Drawable, I think this controls layering

maxrchung:
Closes #77


S2VX.Game/Editor/ToolState/SelectToolState.cs
```diff
@@ -102,7 +102,7 @@ public class SelectToolState : S2VXToolState {
                 foreach (var noteAndTime in SelectedNoteToTime) {
                     var note = noteAndTime.Key;
                     var newTime = GetClosestTickTime(gameTimeAtMouse) + NoteToDragPointDelta[note];
-                    note.EndTime = newTime;
+                    note.UpdateEndTime(newTime);
```
maxrchung, line 105: Here's basically the magic. To update both the Approach and Note together, you call a corresponding Update function inside of Note like this.

PR #97, opened 2020-08-31
78 select note in story

### LINKED ISSUE #78: Select note within story display
- Click on note inside of story to select it

### LINKED ISSUE #84: Dragging selected note when note is not visible on NotesTimeline
Currently trying to drag a selected note when not visible in NotesTimeline causes error
whatsGoingOn

Naranja-Sagged:
Closes #78 
Closes #84


S2VX.Game/Editor/ToolState/SelectToolState.cs
```diff
@@ -35,26 +36,93 @@ public class SelectToolState : S2VXToolState {
             return mouseInXRange && mouseInYRange;
         }
 
+        // https://gamedev.stackexchange.com/a/86784
+        private static Vector2 GetRotatedPoint(Vector2 point, Vector2 center, float rotation) {
+            var translatedX = point.X - center.X;
+            var translatedY = point.Y - center.Y;
+
+            var rotatedX = translatedX * (float)Math.Cos(rotation) - translatedY * (float)Math.Sin(rotation);
```
Xenocidel, line 44: Rather than converting to float after each operation, I think it'd make more sense to do all calculations with higher precision and then convert to float at the very end to avoid compounding any rounding errors in between.

S2VX.Game/Editor/ToolState/SelectToolState.cs
```diff
@@ -35,26 +36,93 @@ public class SelectToolState : S2VXToolState {
...
+            var rotatedY = translatedX * (float)Math.Sin(rotation) + translatedY * (float)Math.Cos(rotation);
+
+            return new Vector2(rotatedX + center.X, rotatedY + center.Y);
+        }
+
+        private bool MouseIsOnNote(Vector2 mousePos, Note note) {
```
Xenocidel, line 50: I prefer naming bool functions with Is at the beginning so you can more easily identify them (same with "MouseIsOnTimelineNote")

S2VX.Game/Editor/ToolState/SelectToolState.cs
```diff
@@ -67,7 +135,7 @@ public class SelectToolState : S2VXToolState {
 
             foreach (var noteAndTime in SelectedNoteToTime) {
                 var note = noteAndTime.Key;
-                if (MouseIsOnNote(mousePos, Editor.NotesTimeline.NoteToTimelineNote[note])) {
+                if (Editor.NotesTimeline.NoteToTimelineNote.ContainsKey(note) && MouseIsOnTimelineNote(mousePos, Editor.NotesTimeline.NoteToTimelineNote[note])) {
```
Xenocidel, line 138: This line is >150 chars

S2VX.Game/Editor/ToolState/SelectToolState.cs
maxrchung: Agreed. In general, there is a convention for boolean functions to be phrased as a yes or no question: https://stackoverflow.com/questions/1370840/naming-conventions-what-to-name-a-method-that-returns-a-boolean

S2VX.Game/Editor/ToolState/SelectToolState.cs
maxrchung: I think limiting line length is fine, but shouldn't be fixed through review. I made issue #105 to add this package https://github.com/menees/Analyzers to check for line length.

S2VX.Game/Editor/ToolState/SelectToolState.cs
```diff
@@ -67,7 +125,7 @@ public class SelectToolState : S2VXToolState {
 
             foreach (var noteAndTime in SelectedNoteToTime) {
                 var note = noteAndTime.Key;
-                if (MouseIsOnNote(mousePos, Editor.NotesTimeline.NoteToTimelineNote[note])) {
+                if (Editor.NotesTimeline.NoteToTimelineNote.ContainsKey(note) && IsMouseOnTimelineNote(mousePos, Editor.NotesTimeline.NoteToTimelineNote[note])) {
```
maxrchung, line 128: I dislike seeing long references like this used multiple times. It can contribute to excessive line length as well. In general, if you find yourself using long chains like this `Editor.NotesTimeline.NoteToTimelineNote`, I think it would easier on the eyes to just save a local variable `noteToTimelineNote` and reference that.

PR #132, opened 2020-09-10
130 S2VXUtils testing and bug fix

Xenocidel:


maxrchung (approved):
My ReversibleStackTests have a test structure somewhat similar to this: [[UnitOfWork_StateUnderTest_ExpectedBehavior]](https://stackoverflow.com/a/1594049/13183186). I don't include the UnitOfWork because I think it's obvious that Reversibles are being tested.

I think your test names are on the simpler side but sufficient, so I approved. You can consider adopting mine, the link's, or looking at other conventions to see what works best and for us.

Naranja-Sagged (approved):

PR #135, opened 2020-09-14
111 note coord undo redo

### LINKED ISSUE #111: Add undo redo for note endtime and coordinate changes
Follow up to #101 for modifying endtime and coordinates.

There is an issue where the delta between note endtime updates is too small, so some work may need to be done to remember old/new values

Naranja-Sagged:
Closes #111


S2VX.Game/Editor/ToolState/SelectToolState.cs
```diff
@@ -212,7 +217,35 @@ public class SelectToolState : S2VXToolState {
             }
         }
 
-        public override void OnToolDragEnd(DragEndEvent _) => ToDrag = SelectToolDragState.None;
+        public override void OnToolDragEnd(DragEndEvent _) {
```
maxrchung, line 220: Use _ for variable name only if the variable is not used, see: https://docs.microsoft.com/en-us/dotnet/csharp/discards

It's general naming convention to use _ when a parameter is necessary for some convention function, but don't use the variable. Sometimes you may see it with a fuller name to give some context, e.g. `OnToolDragEnd(DragEndEvent _event)`.

S2VX.Game/Editor/ToolState/SelectToolState.cs
```diff
@@ -173,26 +177,27 @@ public class SelectToolState : S2VXToolState {
             return gameTime + (leftOffset <= rightOffset ? -leftOffset : rightOffset);
         }
 
+        private double GetGameTimeAtMouse(Vector2 localMousePos) {
+            var tickBarNoteSelections = Editor.NotesTimeline.TickBarNoteSelections;
+            var mousePosX = ToSpaceOfOtherDrawable(ToLocalSpace(localMousePos), tickBarNoteSelections).X;
+            // temp until NoteTimeline Scroll on drag is implemented
```
maxrchung, line 183: What does this comment mean?

S2VX.Game/Editor/ToolState/SelectToolState.cs
```diff
@@ -173,26 +177,27 @@ public class SelectToolState : S2VXToolState {
...
-                        var relativeMousePosX = mousePosX / tickBarNoteSelections.DrawWidth;
-                        var gameTimeDeltaFromMiddle = (relativeMousePosX - 0.5f) * Editor.NotesTimeline.SectionLength * NotesTimeline.SecondsToMS;
-                        var gameTimeAtMouse = Time.Current + gameTimeDeltaFromMiddle;
-
-                        var selectedNoteToTimeCopy = new Dictionary<Note, double>(SelectedNoteToTime);
-                        foreach (var noteAndTime in SelectedNoteToTime) {
-                            var note = noteAndTime.Key;
+                        var gameTimeAtMouse = GetGameTimeAtMouse(e.ScreenSpaceMousePosition);
+
+                        foreach (var note in SelectedNoteToTime.Keys.ToList()) {
```
maxrchung, line 196: Don't think you need ToList(), SelectedNoteToTime.Keys is an IEnumerable, and you can enumerate through this without having to turn it into a list. Should only convert it to List if you need more List functionalities.

S2VX.Game/Editor/ToolState/SelectToolState.cs
Naranja-Sagged: Osu!'s NoteTimeline scrolls when a note is dragged to the edges. I added a card in Tasks a few hours ago so I can maybe handle this in the future. Right now I'm just clamping mousePox within the NoteTimeline.

S2VX.Game/Editor/ToolState/SelectToolState.cs
Naranja-Sagged: I'm using .ToList() to make a copy of the keys so it doesn't throw an error because I'm editing the dictionary I'm iterating over. I was using a temp dict to do this before but I changed it to try to get something else working

S2VX.Game/Editor/ToolState/SelectToolState.cs
Naranja-Sagged: Oh I forgot about that, Changed

S2VX.Game/Editor/ToolState/SelectToolState.cs
maxrchung: Sounds good

S2VX.Game/Editor/ToolState/SelectToolState.cs
maxrchung: That's right I remember that, sounds good

maxrchung (approved):

Xenocidel (approved):

PR #147, opened 2020-09-17
Added camera tool snapping; changed default snapping

Xenocidel:


S2VX.Game/Editor/ToolState/CameraToolState.cs
```diff
@@ -205,10 +226,13 @@ public class CameraToolState : S2VXToolState {
                 new Vector3(relativeNewPosition.X, relativeNewPosition.Y, 0)
             );
             var direction = cross.Z > 0 ? 1 : -1;
-            var radiansBetween = direction * Math.Acos(dot / magnitude);
+            var radiansBetween = direction * Math.Acos(Math.Clamp(dot / magnitude, -1, 1));
```
maxrchung, line 229: Can we add a small comment for the floating point error here? Think we should give context for weirder code so we know why we're doing this down the line

maxrchung (approved):

Naranja-Sagged (approved):

Issue #121, opened 2020-09-07
Test play score system and score feedback


maxrchung:
- Research - discovery into how to best do judgments/timings
- Show some debug UI for score information
- Be able to hide debug UI
- Maybe include accuracy/track progress
- Maybe put [Cached] Score system into PlayScreen and [Resolved] inside of GameNote

PR #156, opened 2020-09-26
123 play hitsound on hit

### LINKED ISSUE #123: Figure out how to play hitsounds when note is clicked
- We can just use a hardcoded hitsound for now
- Probably need to put hitsound into Resources/Samples and then reference it through AudioManager
- Maybe some kind of Square Wave sound? Since this is a Square Game.
- Add hitsound to EditorNote

maxrchung:
Closes #123 

- Hit sound plays for editor note when it is expected to be played. I have some simple flagging logic that prevents duplicate sounds from being played.
- Play mode note click triggers a miss sound. This is temporary. I figure whoever's doing more of the logic with note clicking or score tracking can set this up more appropriately.
- I reworked Volume so that it sets the global Volume instead of just for Track. I'm hooking onto osu's default configuration system for this, so you should see your volume changes automatically saved inside of a file similar to this one: C:\Users\Wax Chug da Gwad\AppData\Roaming\S2VX\framework.ini

Drafting this until #121 PR is merged - I need some changes, and I'll integrate Miss and Hit sounds properly into GameNote


Naranja-Sagged (approved):

Xenocidel:
Starting Play in the middle of the song after some notes registers them as misses and plays the miss sound for each.

maxrchung:
> Starting Play in the middle of the song after some notes registers them as misses and plays the miss sound for each.

This is because Delete() is being called in the Update() function for each GameNote. I'll look into this and maybe provide a fix.

maxrchung:
I fixed this by explicitly removing some notes from Story before it starts playing. This prevents hitsounds from being played and, another issue that would have come up, score from being calculated.

S2VX.Game/Story/S2VXStory.cs
```diff
@@ -87,6 +88,20 @@ public class S2VXStory : CompositeDrawable {
             Approaches.RemoveApproach(note);
         }
 
+        // Before starting the Story in the PlayScreen, we want to explicitly
+        // remove GameNotes up to some certain track time. This is so that we
+        // won't hear Miss hitsounds and prematurely calculate score.
+        public void RemoveNotesUpTo(double trackTime) {
+            while (Notes.Children.Count > 0 && Notes.Children.First().EndTime < trackTime) {
```
Xenocidel, line 95: I still see the issue of notes being counted as Misses in the case where you still have notes after trackTime. I breakpointed and investigated Notes.Children(). They are not ordered by EndTime, seems like they are arbitrarily ordered.

Xenocidel (changes requested):

Xenocidel (approved):

PR #158, opened 2020-09-28
124 click one note

### LINKED ISSUE #124: Restrict to clicking top note
- Look into setting up hot keys and on click/on key handlers
- Some kind of alpha? check to prevent clicks on disappeared notes. For optimization, may need to look into destroy notes so that handlers aren't just ran for no reason.
- Make the note disappear when you click it

Naranja-Sagged:
Closes #124


S2VX.Game/Story/Note/GameNote.cs
```diff
@@ -1,11 +1,127 @@
...
+        }
+
+        private void RecordMiss() {
+            var missThreshold = Story.Notes.MissThreshold;
+            TimingError = missThreshold;
+            Score.Value += missThreshold;
+            Delete();
+        }
+
+        private GameNote GetEarliestClickableNote() {
```
maxrchung, line 37: As discussed, we should look in a better sorted list solution rather than run this every click.

maxrchung (changes requested):

Xenocidel:
Please rebase/merge from master

S2VX.Game/Story/Note/GameNote.cs
maxrchung: I sorted Notes, pull master for the latest changes.

S2VX.Game/Story/Note/GameNote.cs
```diff
@@ -52,40 +54,72 @@ public class GameNote : S2VXNote {
...
 
         protected override bool OnMouseDown(MouseDownEvent e) {
-            if (IsClickable()) {
-                ScoreInfo.AddScore(Math.Abs(TimingError));
-                Delete();
+            if (IsClickable() && !BlockClicking) {
+                ClickNote();
             }
+
+            CheckAndResetClicking();
```
Xenocidel, line 86: For some reason even though it looks like the same logic, the reset blocking doesn't work for clicks

Xenocidel (changes requested):

S2VX.Game/Story/Note/GameNote.cs
maxrchung: I can confirm, clicks do not work. If you think about it, it does make sense, CheckAndResetClicking() only resets the BlockClicking if you are at the last note. No matter where your cursor is, KeyDown always triggers for all notes, but MouseDown may not trigger on the last note if your mouse is not over the note.

S2VX.Game/Story/Note/GameNote.cs
```diff
@@ -52,40 +54,72 @@ public class GameNote : S2VXNote {
...
-        // Notes are clickable if they are visible on screen and not missed
+        // Notes are clickable if they are visible on screen, not missed, and is the earliest note
         private bool IsClickable() {
             var time = Time.Current;
-            // Limit timing error to be +/- MissThreshold (though it will never be >= MissThreshold since RecordMiss would have already run) 
+            // Limit timing error to be +/- MissThreshold (though it will never be >= MissThreshold since RecordMiss would have already run)
             TimingError = (int)Math.Round(Math.Clamp(time - EndTime, -MissThreshold, MissThreshold));
-            return TimingError <= MissThreshold && Alpha > 0;
+            var inMissThreshold = TimingError <= MissThreshold && Alpha > 0;
+            var earliestNote = (GameNote)Story.Notes.Children[^1];
```
maxrchung, line 63: Don't think you need the GameNote cast, leaving it as the default S2VXNote should be fine. I believe an == check by default for reference?

S2VX.Game/Story/Note/GameNote.cs
```diff
@@ -52,40 +54,72 @@ public class GameNote : S2VXNote {
...
+
+        private void ClickNote() {
+            ScoreInfo.AddScore(Math.Abs(TimingError));
+            Delete();
+            BlockClicking = true;
+        }
+
+        // Unblocks clicking on the last call of OnMouseDown / OnKeyDown
+        private void CheckAndResetClicking() {
+            if (Story.Notes.Children.Count > 0 && this == Story.Notes.Children[0]) {
```
maxrchung, line 76: Instead of `Children[^1]` and `Children[0]`, you might want to use the more idiomatic `Children.Last()` and `Children.First()` functions. I don't think you need the `Children.Count > 0` check either because if this function triggers there should always be a `Note`. If you'd like to be safe (and also my preference), you can use the `Children.LastOrDefault()` and `Children.FirstOrDefault()` functions. These functions will return `null` if First or Last do not exist, and you can compare that directly with `this`.

S2VX.Game/Story/Note/GameNote.cs
```diff
@@ -52,40 +54,72 @@ public class GameNote : S2VXNote {
...
+            CheckAndResetClicking();
             return false;
         }
 
         protected override bool OnKeyDown(KeyDownEvent e) {
-            if (IsClickable()) {
+            if (IsClickable() && IsHovered && !BlockClicking) {
                 switch (e.Key) {
                     case Key.Z:
+                        ClickNote();
```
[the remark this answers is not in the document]
Naranja-Sagged: True. I'll change that

S2VX.Game/Story/Note/GameNote.cs
Naranja-Sagged: Oh ok

S2VX.Game/Story/Note/GameNote.cs
```diff
@@ -79,17 +90,21 @@ public class GameNote : S2VXNote {
...
                     default:
                         break;
                 }
             }
+
             return false;
         }
 
         protected override void Update() {
+            if (ShouldBeRemoved) {
```
maxrchung, line 104: Can we leave a comment in the code explaining why this flag is used like this. I think this is a pretty odd usage that should have some documentation/notation explaining what's going on.

maxrchung (commented):
I can confirm

maxrchung (approved):

Xenocidel (approved):

PR #179, opened 2020-10-09
137 Add tests for S2VXCommmand

Xenocidel:


Naranja-Sagged (approved):

S2VX.Game.Tests/CommandTests.cs
```diff
@@ -0,0 +1,150 @@
...
+            var input = "ApproachesDistance|0.1|0.2|None|0.1|1.0";
+            var expected = new ApproachesDistanceCommand() {
+                StartTime = 0.1f,
+                EndTime = 0.2f,
+                Easing = Enum.Parse<Easing>("None"),
+                StartValue = 0.1f,
+                EndValue = 1.0f,
+            };
+            var actual = (ApproachesDistanceCommand)S2VXCommand.FromString(input);
+            var tolerance = 0.001;
```
maxrchung, line 23: I think this can be extracted out to a reusable static variable

S2VX.Game.Tests/CommandTests.cs
```diff
@@ -0,0 +1,150 @@
...
+            Assert.AreEqual(expected.StartTime, actual.StartTime, tolerance);
+            Assert.AreEqual(expected.EndTime, actual.EndTime, tolerance);
+            Assert.AreEqual(expected.Easing, actual.Easing);
+            Assert.AreEqual(expected.StartValue, actual.StartValue);
+            Assert.AreEqual(expected.EndValue, actual.EndValue);
+        }
+
+        [Test]
+        public void ValueAt_Double_0StartTime0DurationUsesStartValue() {
```
maxrchung, line 70: I think these ValueAt tests are better off in a separate InterpolationTests.cs file, since it looks like it's the Interpolation that is being tested, not Commands.

S2VX.Game.Tests/CommandTests.cs
```diff
@@ -0,0 +1,150 @@
...
+            var inputStartValue = Vector2.Zero;
+            var inputEndValue = Vector2.One;
+            var inputStartTime = 1.0f;
+            var inputEndTime = 1.0f;
+            var inputEasing = Easing.None;
+            var expected = inputStartValue;
+            var actual = Interpolation.ValueAt(inputCurrentTime, inputStartValue, inputEndValue, inputStartTime, inputEndTime, inputEasing);
+            Assert.AreEqual(expected, actual);
+        }
+
```
maxrchung, line 148: Just to ensure my sanity, can we also add a few cases for great than 0 Duration? My assumption is that all of these should use the end value, but I'd like the tests to confirm this.

S2VX.Game.Tests/CommandTests.cs
Xenocidel: I'll make it a const

S2VX.Game.Tests/CommandTests.cs
Xenocidel: Now that I think about it, since it's an osu! framework command, we shouldn't need to have unit tests for them at all. Should I just remove them?

maxrchung: As discussed, looks like osu! framework doesn't have a lot of unit tests for these edge cases, so we'll keep our tests for ourselves

maxrchung (approved):

PR #181, opened 2020-10-09
180 Fix camera preview when trying to create an overlapping command

Xenocidel:


Xenocidel:
This also addresses a bug where the camera preview is only visible once (i.e. you couldn't go back and see what the movement looks like). This is done by actually writing the command to the command panel.

Xenocidel:
Locks are necessary on Camera (please use SetPosition, SetScale, and SetRotation) so that preview is not blocked by the constant updating of pre-existing commands.

Naranja-Sagged (approved):

maxrchung:
Locks feel a bit weird to me, but I don't see other ways to really make this work with what we've got now. I think we'll discuss this a bit more during our meeting, but overall I'm approved.

maxrchung (approved):

PR #187, opened 2020-10-12
172 code structure for song selection screen

### LINKED ISSUE #172: Code structure for Song Selection screen
Initial screen in Game should be Song Selection, you can push a PlayScreen or EditorScreen from there

Proposed file structure:
- Stories are always the last leaf nodes
- A folder that contains Stories cannot also be a Story itself
Stories
----Camellia Remixes
--------Camellia Remix
------------story.json
------------Camellia Remix.mp3
--------Camellia Remix2
------------story.json
------------Camellia Remix.mp3
----Camellia Singles
--------Camellia Single
------------story.json
------------Camellia Remix.mp3

maxrchung:
Also made these miscellaneous changes that I thought were beneficial when play testing and going through code:
- Reverse snap divisor mouse wheel behavior so that mouse wheel down decreases the value and mouse up increases value
- G to play from editor screen so it's easier to initiate test play
- Optimize RemoveNotesUpTo to use `ClearInternal`

Closes #172

### DIFF
--- a/S2VX.Game/Editor/EditorScreen.cs
+++ b/S2VX.Game/Editor/EditorScreen.cs
@@ -22,8 +22,7 @@ namespace S2VX.Game.Editor {
     public class EditorScreen : Screen {
         public Vector2 MousePosition { get; private set; } = Vector2.Zero;
 
-        [Cached]
-        private S2VXStory Story { get; set; } = new S2VXStory();
+        private S2VXStory Story { get; set; }
 
         public CommandPanel CommandPanel { get; } = new CommandPanel();
         private bool IsCommandPanelVisible { get; set; }
@@ -61,6 +60,15 @@ namespace S2VX.Game.Editor {
 
         [Resolved]
         private ScreenStack Screens { get; set; }
+
+        // We need to explicitly cache dependencies like this so that we can
+        // recache an EditorScreen whenever a new one is pushed
+        protected override IReadOnlyDependencyContainer CreateChildDependencies(IReadOnlyDependencyContainer parent) {
+            var dependencies = new DependencyContainer(parent);
+            dependencies.Cache(this);
+            dependencies.Cache(Story = new S2VXStory());
+            return dependencies;
+        }
 
         [BackgroundDependencyLoader]
         private void Load() {
@@ -103,7 +111,8 @@ namespace S2VX.Game.Editor {
                             {
                                 new MenuItem("Refresh (Ctrl+R)", ProjectRefresh),
                                 new MenuItem("Save (Ctrl+S)", ProjectSave),
-                                new MenuItem("Play (P)", ProjectPlay)
+                                new MenuItem("Play (P)", ProjectPlay),
+                                new MenuItem("Quit (Ctrl+Q)", ProjectQuit),
                             }
                         },
                         new MenuItem("Edit")
@@ -199,6 +208,11 @@ namespace S2VX.Game.Editor {
                 case Key.P:
                     if (e.ControlPressed) {
                         ProjectPlay();
+                    }
+                    break;
+                case Key.Q:
+                    if (e.ControlPressed) {
+                        ProjectQuit();
                     }
                     break;
                 case Key.R:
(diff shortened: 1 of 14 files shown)


S2VX.Game/Story/Note/Notes.cs
```diff
@@ -36,6 +37,23 @@ public class Notes : CompositeDrawable {
...
+            int validIndex;
+            for (validIndex = 0; validIndex < Children.Count; ++validIndex) {
+                if (Children[validIndex].EndTime < trackTime) {
+                    break;
+                }
+            }
+
+            var newNotes = Children.Take(validIndex).ToList();
+            Children = newNotes;
+            ClearInternal(false);
```
Xenocidel, line 53: Do we need the old notes list? What does disposeChildren = false do for us here?

S2VX.Game/Editor/EditorScreen.cs
```diff
@@ -325,6 +332,8 @@ public class EditorScreen : Screen {
             Story.Save(@"../../../story.json");
         }
 
+        private void ProjectQuit() => this.Exit();
```
Xenocidel, line 335: Pressing escape while the song is playing causes the song to continue playing

Xenocidel (changes requested):

S2VX.Game/SongSelection/SongSelectionScreen.cs
```diff
@@ -0,0 +1,37 @@
...
+using osu.Framework.Graphics;
+using osu.Framework.Graphics.Containers;
+using osu.Framework.Graphics.UserInterface;
+using osu.Framework.Screens;
+using osuTK;
+using S2VX.Game.Editor;
+
+namespace S2VX.Game.Play {
+    public class SongSelectionScreen : Screen {
+        private static Vector2 InputSize { get; } = new Vector2(100, 30);
```
Xenocidel, line 11: General question: is there any particular reason to use `readonly` over `const` in cases where we're just setting the value at the start and never changing it ever again?

Xenocidel: Oh right, this is calling a constructor and so `const` won't work because it's not evaluated at compile-time. This can be set to `readonly` though.

S2VX.Game/Story/Note/Notes.cs
maxrchung: We need to call ClearInternal(false) so that Notes are not disposed of before reuse. The main problem lies with the next line: InternalChildren = Childre`. When you assign InternalChildren a new Drawable list, it first invalidates all the existing Drawables before setting the new Drawables. This is a problem for `RemoveNotesUpTo` because some of the same Notes are being used before and after InternalChildren is set.

We need to call ClearInternal(false) so that we first clear the InternalChildren without invalidating the Drawables, then we can safely set InternalChildren without invalidation errors.

S2VX.Game/Editor/EditorScreen.cs
maxrchung: Added Track to EditorScreen's InternalChildren so it is removed properly when EditorScreen exits

S2VX.Game/SongSelection/SongSelectionScreen.cs
maxrchung: I don't think you can use `readonly` with properties. 
image

I've essentially done the same thing with properties by only allowing a getter.

Xenocidel (approved):

S2VX.Game/SongSelection/SongSelectionScreen.cs
Naranja-Sagged, line 11: Wouldn't TextSize be a more fitting variable name in this case since the user would not be inputting anything?

maxrchung: True, I'll change it to BtnSize tomorrow since it has to do with buttons

maxrchung: Merged and updated to BtnSize

Naranja-Sagged (approved):

PR #190, opened 2020-10-17
170 Song Selection (and IconButton)

Xenocidel:
Also creates a new IconButton class. This IconButton can use either a built-in FontAwesome icon **(Icon = )** or an image from TextureStore **(TextureName = )**. Maybe a future improvement could be to move the text label outside the button, but currently I'm not using it to display any text.
image
image


S2VX.Game/SongSelection/UserInterface/BorderOuterBox.cs
```diff
@@ -0,0 +1,30 @@
...
+
+        // Is called by InnerBox when mouse hovers over InnerBox (and thus should not be considered over OuterBox)
+        public void LoseHover() => Colour = Color4.White;
+
+        protected override bool OnHover(HoverEvent e) {
+            Colour = Color4.LightGray;
+            return false;
+        }
+
+        protected override void OnHoverLost(HoverLostEvent e) => LoseHover();
```
Naranja-Sagged, line 22: Instead of having the LoseHover() function, wouldn't it work if you just did everything in this function? The LoseHover call in BorderInnerBox is also unnecessary I believe, since OnHoverLost would be called in here.

S2VX.Game/IconButton.cs
```diff
@@ -0,0 +1,152 @@
...
+        public double DisabledFadeDuration { get; set; } = 200;
+
+        protected Box Hover { get; set; }
+        protected Box Background { get; set; }
+        protected Drawable Image { get; set; }
+
+        [BackgroundDependencyLoader]
+        private void Load(TextureStore store) {
+            if (!Icon.Equals(new IconUsage())) {    // Equivalent to != null
+                var iconScale = 0.8f;
```
maxrchung, line 86: I think this is something we probably want to make configurable

S2VX.Game/IconButton.cs
```diff
@@ -0,0 +1,152 @@
...
+            return base.OnHover(e);
+        }
+
+        protected override void OnHoverLost(HoverLostEvent e) {
+            base.OnHoverLost(e);
+
+            Hover.FadeOut(HoverFadeDuration);
+        }
+
+        //private void EnabledChanged(ValueChangedEvent<bool> e) =>
```
maxrchung, line 149: Remove comment

S2VX.Game/SongSelection/SongSelectionScreen.cs
```diff
@@ -1,37 +1,150 @@
...
+        private (bool, string, string) DirectoryContainsStory(string dir) {
+            var story = Storage.GetFiles(dir, "*.s2ry");
+            var song = Storage.GetFiles(dir, "audio.mp3");
+            return (story.Count() == 1 && song.Count() == 1, story.FirstOrDefault(), song.FirstOrDefault());
+        }
+
+        private bool DirectoryContainsDirectories(string dir) => Storage.GetDirectories(dir).Any();
+
+        // Go up one level by exiting and thus popping ourself out from the ScreenStack
+        public void DoExit() {
```
maxrchung, line 51: Can you just override IScreen's OnExiting instead? I think this will also remove the need to pass SongSelectionScreen everywhere.

maxrchung (commented):
UI looking good

S2VX.Game/SongSelection/UserInterface/BorderOuterBox.cs
Xenocidel: You're right, so long as I capture hover inside BorderInnerBox (return true), it'll act as if hover is lost in BorderOuterBox.

S2VX.Game/SongSelection/SongSelectionScreen.cs
```diff
@@ -1,37 +1,150 @@
...
         private void Load() {
-            BtnEdit.Action = () => this.Push(new EditorScreen());
-            BtnPlay.Action = () => this.Push(new PlayScreen(false));
-            InternalChild = new FillFlowContainer {
-                Direction = FillDirection.Horizontal,
-                Children = new Drawable[]
-                {
-                    BtnEdit,
-                    BtnPlay
+            Audio.Samples.Get("menuhit").Play();
```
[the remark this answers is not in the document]
Xenocidel: This was done so that the sound would play when a new SongSelectionScreen is loaded

S2VX.Game/IconButton.cs
Xenocidel: Will do

S2VX.Game/Editor/EditorScreen.cs
```diff
@@ -20,6 +22,24 @@
...
     public class EditorScreen : Screen {
+        private string AudioDir { get; set; }
+        private StorageBackedResourceStore CurLevelResourceStore { get; set; }
+        private string CurSelectionPath { get; set; }
+        private string StoryDir { get; set; }
+        private string FullStoryDir { get; set; }
+        public EditorScreen(string curSelectionPath, string storyDir, StorageBackedResourceStore curLevelResourceStore, string audioDir) {
+            CurSelectionPath = curSelectionPath;
+            StoryDir = storyDir;
+            AudioDir = audioDir;
```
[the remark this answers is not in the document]
Xenocidel: Will rename to path

S2VX.Game/IconButton.cs
Xenocidel: Will do

S2VX.Game/SongSelection/SongSelectionScreen.cs
Xenocidel: If I do that then I'll need to move the logic for checking for root level to BorderOuterBox's OnClick. Is that preferred?

maxrchung: The OnExit function says that it will prevent exit if you return true, can't you use this? 

Isn't the current selection screen always on top? Can you not do some kind of generic screen poptop to remove the top screen?

S2VX.Game/SongSelection/SongSelectionScreen.cs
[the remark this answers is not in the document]
maxrchung: I found it a bit odd that a sound effect plays at the initial load of the game. To me, the sound effect isnt associated so much with a new screen being loaded, but when a story/folder square is clicked

S2VX.Game/SongSelection/SongSelectionScreen.cs
Xenocidel: Thanks for pointing out the return true behavior. Will do.

S2VX.Game/SongSelection/SongSelectionScreen.cs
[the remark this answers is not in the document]
Xenocidel: Will do

maxrchung (approved):

S2VX.Game/Play/PlayScreen.cs
```diff
@@ -12,16 +14,32 @@
...
+        private string StoryDir { get; set; }
+        private string FullStoryDir { get; set; }
+        public PlayScreen(bool isUsingEditorSettings, string curSelectionPath, string storyDir,
+            StorageBackedResourceStore curLevelResourceStore, string audioDir) {
 
-        public PlayInfoBar PlayInfoBar { get; private set; } = new PlayInfoBar();
+            IsUsingEditorSettings = isUsingEditorSettings;
+            CurSelectionPath = curSelectionPath;
+            StoryDir = storyDir;
+            AudioDir = audioDir;
```
maxrchung, line 28: Should rename so it's not StoryDir and AudioDir

Xenocidel: Done

Naranja-Sagged (approved):

PR #217, opened 2020-10-31
189 Add HoldNote objects

Xenocidel:


S2VX.Game/Story/Note/EditorHoldNote.cs
```diff
@@ -0,0 +1,73 @@
...
+
+            switch (NumHitSounds) {
+                case 2:
+                    if (time >= HitTime) {
+                        --NumHitSounds;
+                        Hit.Play();
+                    }
+                    break;
+                case 1:
+                    if (time >= EndTime) {
```
Xenocidel, line 54: I realize I can collapse this down into one

Naranja-Sagged (changes requested):
HoldNotes get stuck on the left side of the notesTimeline when current time is past the HoldNote's endTime
image

Naranja-Sagged (changes requested):
Pressing another input and releasing that same input too early will count as a miss even if user is still holding down the original input they initially pressed on the HoldNote. I think having at least one input held down should prevent a miss, like how you can hold and switch inputs during a slider in osu!.

S2VX.Game/Editor/Containers/NotesTimeline.cs
```diff
@@ -259,21 +259,43 @@ public class NotesTimeline : CompositeDrawable {
...
-                        Height = TimelineNoteHeight,
-                        X = (float)relativePosition,
-                        Y = 0.2f,
-                        Anchor = Anchor.TopLeft,
-                    };
+                RelativeBox visibleNote = null;
+                var relativePosition = (note.HitTime - lowerBound) / sectionDuration;
+
+                if (note is EditorHoldNote holdNote) {
+                    if (lowerBound <= holdNote.HitTime || holdNote.EndTime <= upperBound) {
```
Naranja-Sagged, line 271: The HoldNote getting stuck on the left side of NotesTimeline probably has something to do with the logic here. I tried changing the || to && but the HoldNote disappears too quickly Lol

maxrchung:
Pausing my review until Alex's issues are addressed. I don't think I have anything else to add though, code structure looks fine.

S2VX.Game/Editor/Containers/NotesTimeline.cs
Xenocidel: Fixed

Xenocidel:
> Pressing another input and releasing that same input too early will count as a miss even if user is still holding down the original input they initially pressed on the HoldNote. I think having at least one input held down should prevent a miss, like how you can hold and switch inputs during a slider in osu!.

I tried changing IsBeingHeld to be NumInputsHeld but KeyDownEvent gets called multiple times if you hold down a key (key repeats as if it were a text editor). Need to figure that out.

maxrchung:
> Pressing another input and releasing that same input too early will count as a miss even if user is still holding down the original input they initially pressed on the HoldNote. I think having at least one input held down should prevent a miss, like how you can hold and switch inputs during a slider in osu!.

Aaron and I discussed switching inputs in the middle of a hold/slider. I think the key you start a hold/slider with should be the same key you release with. This way also seems a bit easier to implement compared to allowing switching inputs in the middle of the hold/slider.

maxrchung (approved):

Naranja-Sagged (approved):

PR #219, opened 2020-11-01
Go to start when refreshing or opening story

maxrchung:
Pending @Naranja-Sagged's attempt to look into a fix, here's my workaround where refreshing or opening a story goes to the start of the song


maxrchung:
Closing to remake with right name and closing

PR #221, opened 2020-11-01
127 indicate error while adding command

### LINKED ISSUE #127: Indicate error when adding command in Command Panel
As a story writer, I want my game not to crash when I try and add an invalid command in the command panel.

This can happen because of invalid input type, e.g. entering invalid vector2 string.

Just prevent exception/crash: catch the exception and return generic error.

Naranja-Sagged:
Closes #127


S2VX.Game/Editor/Containers/CommandPanel.cs
```diff
@@ -79,7 +85,59 @@ public class CommandPanel : OverlayContainer {
...
+                Width = 0.0025f,
+                Height = 0.04f,
+                Y = .075f,
+                X = -.051f,
+            });
+        }
+
         private void HandleAddClick() {
+            ErrorContainer.Clear();
+            if (!double.TryParse(TxtStartTime.Current.Value, out var _) || !double.TryParse(TxtEndTime.Current.Value, out var _)) {
```
maxrchung, line 137: I don't think we need these `double.TryParse` calls. `S2VXCommand.FromString` already does `double.Parse`.

S2VX.Game/Editor/Containers/CommandPanel.cs
```diff
@@ -79,7 +85,59 @@ public class CommandPanel : OverlayContainer {
...
+                Origin = Anchor.Centre,
+                Colour = Color4.Red,
+                Width = 0.0025f,
+                Height = 0.04f,
+                Y = .075f,
+                X = -.051f,
+            });
+        }
+
         private void HandleAddClick() {
```
maxrchung, line 135: I would try catch this whole block and `Console.WriteLine` the Exception in the catch block. You've mentioned that you have to continue after getting an exception in `S2VXCommand.FromString` but this is fine.

S2VX.Game/Story/Command/S2VXCommand.cs
```diff
@@ -55,11 +55,15 @@ public abstract class S2VXCommand : IComparable<S2VXCommand> {
...
-            command.Easing = Enum.Parse<Easing>(split[3]);
-            return command;
+            try {
+                var command = staticMethod.Invoke(null, new object[] { split }) as S2VXCommand;
+                command.StartTime = double.Parse(split[1], CultureInfo.InvariantCulture);
+                command.EndTime = double.Parse(split[2], CultureInfo.InvariantCulture);
+                command.Easing = Enum.Parse<Easing>(split[3]);
+                return command;
+            } catch (TargetInvocationException) {
+                return null;
```
maxrchung, line 65: I don't think we should put try catch in this function

maxrchung (changes requested):

S2VX.Game/Editor/Containers/CommandPanel.cs
Xenocidel: Why the entire function? Couldn't it just be around the S2VXCommand.FromString part?

S2VX.Game/Story/Command/S2VXCommand.cs
Xenocidel: If we only handle errors when adding new commands from the CommandPanel, then it'd still crash if someone added a command manually by editing the .s2ry right?

Xenocidel, line 65: Console.WriteLine the exception info in catch block

Xenocidel (changes requested):

S2VX.Game/Story/Command/S2VXCommand.cs
maxrchung: Yeah, I would add another try catch block when loading the story. I'm not sure of the best practice for exception handling, but it seems pretty awkward if we try and catch errors in S2VXCommand itself because S2VXCommand wouldn't really know how to properly handle the exception, e.g. rolling back changes or showing the error indicator to user.

S2VX.Game/Editor/Containers/CommandPanel.cs
```diff
@@ -90,8 +145,16 @@ public class CommandPanel : OverlayContainer {
                 $"{TxtEndValue.Current.Value}"
             };
             var join = string.Join("|", data);
-            var command = S2VXCommand.FromString(join);
-            HandleAddCommand(command);
+            try {
+                var command = S2VXCommand.FromString(join);
+                HandleAddCommand(command);
+            } catch (FormatException) {
+                AddErrorIndicator();
```
maxrchung, line 152: Write out the exception in console so we log exactly what's wrong

maxrchung (approved):

Xenocidel (approved):

PR #234, opened 2020-11-07
223 HoldNote tool creation process

### LINKED ISSUE #223: HoldNote tool creation process
As a story maker, I want to create a hold note with a defined start and end using the hold note tool so that I don't have to create the note then adjust the end time later.

* First click sets the position and hit time
* Second click sets the end time
* Tool description should say something like "Click to set hit time" and "Click to set end/release time"

maxrchung:
Closes #223

### DIFF
--- a/S2VX.Game/Editor/ToolState/HoldNoteToolState.cs
+++ b/S2VX.Game/Editor/ToolState/HoldNoteToolState.cs
@@ -28,18 +28,20 @@ namespace S2VX.Game.Editor.ToolState {
             Child = Preview;
         }
 
-        public override bool OnToolClick(ClickEvent _) {
+        public override bool OnToolClick(ClickEvent e) {
             if (!IsRecording) {
                 HitTime = Time.Current;
                 Coordinates = Editor.MousePosition;
             } else {
                 var endTime = Time.Current;
-                var note = new EditorHoldNote {
-                    Coordinates = Coordinates,
-                    HitTime = HitTime,
-                    EndTime = endTime
-                };
-                Editor.Reversibles.Push(new ReversibleAddHoldNote(Story, note));
+                if (endTime > HitTime) {
+                    var note = new EditorHoldNote {
+                        Coordinates = Coordinates,
+                        HitTime = HitTime,
+                        EndTime = endTime
+                    };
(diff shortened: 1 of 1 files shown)


Xenocidel:
This would work with any mouse click, correct? Doesn't matter left or right mouse button

S2VX.Game/Editor/ToolState/HoldNoteToolState.cs
```diff
@@ -24,21 +29,49 @@ public class HoldNoteToolState : S2VXToolState {
...
-                Coordinates = Editor.MousePosition,
-                HitTime = Time.Current,
-                EndTime = Time.Current + 1000    // TODO: #216 be able to change hold duration
-            };
-            Editor.Reversibles.Push(new ReversibleAddHoldNote(Story, note));
+            if (!IsRecording) {
+                HitTime = Time.Current;
+                Coordinates = Editor.MousePosition;
+            } else {
+                var endTime = Time.Current;
```
Xenocidel, line 36: Should we enforce that endTime > HitTime?

maxrchung:
> This would work with any mouse click, correct? Doesn't matter left or right mouse button

That's correct, I'll change this to only left click so it's a bit more natural

S2VX.Game/Editor/ToolState/HoldNoteToolState.cs
maxrchung: That's a good point, I'll see how to address this

maxrchung:
> This would work with any mouse click, correct? Doesn't matter left or right mouse button

Okay, this is a bit weird, but it looks like osu!'s OnClick handler only triggers off of left clicks, so the behavior is intended.

S2VX.Game/Editor/ToolState/HoldNoteToolState.cs
maxrchung: Made a change so that if you click a second time and the end time is <= hit time, the hold note is not created, and you go back to the initial note placement phase. This follows osu!'s way of handling spinners.

Xenocidel (approved):

Naranja-Sagged (approved):

PR #273, opened 2020-12-08
260 Add slider trail to hold note

### LINKED ISSUE #260: Add slider trail to hold note
As a player, I want to see the slider trail for a hold note so that I can know and predict where to move my mouse.
* Update interpolation for editor/game notes
* Snaking in/out up to implementer discretion
* Slider trail mockup:
image

maxrchung:
Closes #260


Naranja-Sagged (approved):

S2VX.Game/Story/Note/Approach.cs
```diff
@@ -29,41 +29,75 @@ public class Approach : CompositeDrawable, IComparable<Approach> {
...
             var thickness = approaches.Thickness;
+            var time = Time.Current;
 
-            var offset = S2VXUtils.Rotate(coordinates - position, rotation) * scale;
-
-            var distance = time < HitTime
-                ? Interpolation.ValueAt(time, approaches.Distance, scale.X / 2, startFadeIn, HitTime)
-                : scale.X / 2;
+            var startTime = HitTime - notes.ShowTime - notes.FadeInTime;
+            var clampedTime = MathHelper.Clamp(time, startTime, HitTime);
```
Xenocidel, line 93: What's the diff between using Math.Clamp and MathHelper.Clamp? We've been using Math.Clamp throughout the codebase thus far

S2VX.Game/Story/Note/EditorNote.cs
```diff
@@ -9,19 +10,48 @@ public class EditorNote : S2VXNote {
         private bool CanHit { get; set; }
         private SampleChannel Hit { get; set; }
 
+        [Resolved]
+        private S2VXStory Story { get; set; }
```
Xenocidel, line 14: Is this needed?

S2VX.Game/Story/Note/GameNote.cs
```diff
@@ -104,22 +104,41 @@ public class GameNote : S2VXNote {
...
+            }
+
             if (time >= HitTime) {
-                // Hold the note at fully visible until after MissThreshold
-                var notes = Story.Notes;
-                var startFadeTime = HitTime + MissThreshold;
-                var endFadeTime = startFadeTime + notes.FadeOutTime;
-                var alpha = Interpolation.ValueAt(time, 1.0f, 0.0f, startFadeTime, endFadeTime);
-                Alpha = alpha;
                 Colour = Color4.Red;
```
Xenocidel, line 140: This will cause the note to remain Red if you start the hold late, is that fine?

S2VX.Game/Story/Note/HoldNote.cs
```diff
@@ -8,5 +14,104 @@ public abstract class HoldNote : S2VXNote {
...
+
+        [Resolved]
+        private ScreenStack Screens { get; set; }
+
+        [BackgroundDependencyLoader]
+        private void Load() => AddInternal(SliderPath);
+
+        private Path SliderPath { get; set; } = new Path() {
+            // Show slider path behind note
+            Depth = 100,
```
Xenocidel, line 26: What is this Depth and magic number?

Xenocidel (changes requested):

Xenocidel:
Lag occurs before/during HoldNote if there is camera movement on screen (seems like more movement -> more lag)

S2VX.Game/Story/Note/GameNote.cs
maxrchung: I think you meant to comment this in GameHoldNote.cs.

This issue is happening in master. In this checkin, I just replicated the behavior.

I don't think this behavior's fine though, as I'm confused if I missed it or think I'm like more punished for hitting notes late than early. I made an issue describing this issue: https://github.com/maxrchung/S2VX/issues/274. We can talk about solutions or visual changes in our next meeting.

S2VX.Game/Story/Note/Approach.cs
maxrchung: This is an interesting one. So Math.Clamp is from .NET's libraries and MathHelper.Clamp is from osuTK, a fork of the Open Toolkit library that "is a fast, low-level C# binding for OpenGL, OpenGL ES and OpenAL." osuTK comes with math utilities probably to help calculate graphics stuff, and one of those is the Clamp function.

It seems like Math.Clamp was not added to .NET until .NET Core 2.0, at least according to this Stack Overflow post: https://stackoverflow.com/a/2683487/13183186

So using either one doesn't really matter since they do the same thing, just being written separately for mostly historic reasons. To be really pedantic about it, I looked at the source code, and I think .NET's code is maybe a bit safer or optimal? I pushed a change to switch out MathHelper.Clamp calls to Math.Clamp.

.NET source:

```csharp
        [MethodImpl(MethodImplOptions.AggressiveInlining)]
        public static double Clamp(double value, double min, double max)
        {
            if (min > max)
            {
                ThrowMinMaxException(min, max);
            }

            if (value < min)
            {
                return min;
            }
            else if (value > max)
            {
                return max;
            }

            return value;
        }
```

osuTK source:

```csharp
        public static int Clamp(int n, int min, int max)
        {
            return Math.Max(Math.Min(n, max), min);
        }
```

maxrchung:
> Lag occurs before/during HoldNote if there is camera movement on screen (seems like more movement -> more lag)

Looking into this

S2VX.Game/Story/Note/EditorNote.cs
Xenocidel: Oh missed that, thought it wasn't being used.

S2VX.Game/Story/Note/HoldNote.cs
Xenocidel: Using ChangeInternalChildDepth may be safer than setting it directly (there's some checking in there to make sure things are loaded and such).
image

maxrchung: I think ChangeInternalChildDepth or ChangeChildDepth is only needed if the following are true:
* You're trying to manipulate a child's Depth with the parent CompositeDrawable or Container object
* You're trying to manipulate the child's Depth after loading

I don't see anything in the logic of ChangeInternalChildDepth that does anything safer as you suggest:

```csharp
            var index = IndexOfInternal(child);
            if (index < 0)
                throw new InvalidOperationException($"Can not change depth of drawable which is not contained within this {nameof(CompositeDrawable)}.");

            internalChildren.RemoveAt(index);
            var aliveIndex = aliveInternalChildren.IndexOf(child);
            if (aliveIndex >= 0) // remove if found
                aliveInternalChildren.RemoveAt(aliveIndex);

            var chId = child.ChildID;
            child.ChildID = 0; // ensure Depth-change does not throw an exception
            child.Depth = newDepth;
            child.ChildID = chId;

            internalChildren.Add(child);
            if (aliveIndex >= 0) // re-add if it used to be in aliveInternalChildren
                aliveInternalChildren.Add(child);
```

It looks like it's just removing a child, setting the child's Depth, then re-adding the child back in. This makes sense when you're changing a child's depth after loading, but we don't need to do that if we're setting a child on initial load. This looks like a waste of work if I were to change HoldNote.cs:

```csharp
        [BackgroundDependencyLoader]
        private void Load() {
            AddInternal(SliderPath);
            ChangeInternalChildDepth(SliderDepth, 100);
        }
```

This would add SliderPath at Depth 0, remove SliderPath, set SliderPath's Depth to 100, and re-add SliderPath into the CompositeDrawable.

maxrchung: I think ChangeInternalChildDepth or ChangeChildDepth is only needed if the following are true:
* You're trying to manipulate a child's Depth with the parent CompositeDrawable or Container object
* You're trying to manipulate the child's Depth after loading

Here's the bulk of the ChangeInternalChildDepth logic:

```csharp
            var index = IndexOfInternal(child);
            if (index < 0)
                throw new InvalidOperationException($"Can not change depth of drawable which is not contained within this {nameof(CompositeDrawable)}.");

            internalChildren.RemoveAt(index);
            var aliveIndex = aliveInternalChildren.IndexOf(child);
            if (aliveIndex >= 0) // remove if found
                aliveInternalChildren.RemoveAt(aliveIndex);

            var chId = child.ChildID;
            child.ChildID = 0; // ensure Depth-change does not throw an exception
            child.Depth = newDepth;
            child.ChildID = chId;

            internalChildren.Add(child);
            if (aliveIndex >= 0) // re-add if it used to be in aliveInternalChildren
                aliveInternalChildren.Add(child);
```

It looks like it removes a child, sets the child's Depth, then re-adds the child back in. This makes sense for safely changing a child's depth after loading, but we don't need to do that if we're setting a child on initial load. This looks like a waste of work if I were to change HoldNote.cs:

```csharp
        [BackgroundDependencyLoader]
        private void Load() {
            AddInternal(SliderPath);
            ChangeInternalChildDepth(SliderDepth, 100);
        }
```

This would add SliderPath with Depth 0, remove SliderPath from the CompositeDrawable, set SliderPath's Depth to 100, and re-add SliderPath back in to the CompositeDrawable.

maxrchung:
> Lag occurs before/during HoldNote if there is camera movement on screen (seems like more movement -> more lag)

From my initial investigation, osu! sliders appear to similarly update vertices every Update() call: https://github.com/ppy/osu/blob/aeb059bcceb9676e9baa8b285d5a6245c70da362/osu.Game.Rulesets.Osu/Objects/Drawables/DrawableSlider.cs#L205

Not sure if osu!'s slider vertices calculation is somehow more optimized, or if there's a separate issue.

Xenocidel (approved):

Issue #126, opened 2020-09-07
Cursor color, rotation, scale


maxrchung:
As a player, I want my cursor to be stylized as a square so it matches the theme of S2VX: Squares.

- Rotate and scale based on grid
- Hardcode color for now

Interesting idea: Try and incorporate #329 indicator colors into cursor, so we don't have extra elements

PR #276, opened 2020-12-11
216 hold note end time

### LINKED ISSUE #216: Change HoldNote endtime in notes timeline
As a story editor, I want to change the duration of the hold note so that I can change the duration without having to recreate the hold note.

* Click and drag the end of the hold note in the notes timeline to change the duration

Xenocidel:
Closes #216 
Cursor to change to <> when hovering over end of HoldNote in NotesTimeline to be done in #126

### DIFF
--- a/S2VX.Game/Story/Note/EditorHoldNote.cs
+++ b/S2VX.Game/Story/Note/EditorHoldNote.cs
@@ -29,6 +29,15 @@ namespace S2VX.Game.Story.Note {
             HitSoundTimes = new List<double>() { HitTime, EndTime };
         }
 
+        public override void UpdateEndTime(double endTime) {
+            if (endTime < HitTime) {
+                return;
+            }
+            EndTime = endTime;
+            ((HoldApproach)Approach).EndTime = EndTime;
+            HitSoundTimes = new List<double>() { HitTime, EndTime };
+        }
+
         public override void UpdateCoordinates(Vector2 coordinates) {
             // Prevent changes at end time or else we'll get some wild interpolation
             if (EndTime - Time.Current <= EditorScreen.TrackTimeTolerance) {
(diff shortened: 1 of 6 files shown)


S2VX.Game/Story/Note/EditorHoldNote.cs
```diff
@@ -29,6 +29,15 @@ public class EditorHoldNote : HoldNote {
             HitSoundTimes = new List<double>() { HitTime, EndTime };
         }
 
+        public override void UpdateEndTime(double endTime) {
+            if (endTime < HitTime) {
+                return;
+            }
+            EndTime = endTime;
+            ((HoldApproach)Approach).EndTime = EndTime;
```
maxrchung, line 37: Can you see if you can do an override of the Approach property for hold notes? This way we don't have to cast it like this. I think you can set Approach to virtual inside of S2VXNote.cs and then override it in HoldNote.cs. For more details: https://stackoverflow.com/a/9238902/13183186

Xenocidel: Is this good?

maxrchung: I think this is okay. Feels kind of odd having 2 properties point to the same thing. I was hoping that we'd be able to somehow elegantly override the same property as a HoldApproach, but don't think that's possible.

maxrchung (approved):

S2VX.Game/Story/Note/EditorHoldNote.cs
Xenocidel: Yeah, not possible to override the same property name with a different type.

Naranja-Sagged (approved):

PR #289, opened 2020-12-19
272 create release

### LINKED ISSUE #272: Create release to GitHub
As a developer, I want to release our 1.0 directly to GitHub so it can be accessible easily to people.

* People should be able to download it from GitHub and run the game.
* May want to potentially look into automatic releases (with tagging?) or something triggered with pipelines, not sure

* Using Git Flow system of master, dev, and feature branches. 
* Using Release Drafter to automatically draft release notes.
* Adding a S2VX zip with the sample Stories folder

image

### LINKED ISSUE #290: Windows pipeline has issue with installing dependencies
As a developer, I want the Windows pipeline to pass so that we can verify that our project is working for Windows systems.

dotnet restore fails for Windows pipeline. It reports an issue where it tries to get packages locally rather than through NuGet. See issue tracking problem: https://github.com/actions/setup-dotnet/issues/155

Remember to reenable CI for branch protection check

maxrchung:
Closes #272
Closes #290 

See wiki for how new release flow works: https://github.com/maxrchung/S2VX/wiki/Release-Flow


.github/workflows/validate-pull-request.yml
```diff
@@ -0,0 +1,30 @@
...
+  validate-pull-request:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Release cannot be source branch
+      if: ${{ github.head_ref == 'release' }}
+      run: exit error
+
+    - name: Master can only merge to release
+      if: ${{ github.head_ref == 'master' && github.base_ref != 'release' }}
```
Xenocidel, line 21: Should we have the converse as well? If we're merging to release it has to come from master

.github/workflows/continuous-integration.yml
```diff
@@ -1,17 +1,20 @@
-name: Continuous Integration
+name: Run continuous integration
 
 on:
   push:
     branches-ignore:
-      - master
```
Xenocidel, line 6: Do we want CI to run on merges into master? It would have already run on the feature branch, but if we're concerned about potential integration failures then I suppose it makes sense.

Xenocidel (changes requested):

.github/workflows/validate-pull-request.yml
maxrchung: Great catch. I thought this rule would take care of both cases, but thinking about it more realized that isn't the case.

.github/workflows/continuous-integration.yml
maxrchung: I think you're right and prefer to not have to do redundant work. There shouldn't be a point in running the CI twice.

Xenocidel (approved):

Naranja-Sagged (approved):

PR #338, opened 2021-01-19
Remove approach flicker

### LINKED ISSUE #317: Remove approach flicker
As a story editor, I want the approach to be completely gone when the approach thickness is set to 0.

- Right now, you can see some approach flickering

maxrchung:
Closes #317 

Also removed some using statements that I got warnings for

### DIFF
--- a/S2VX.Game.Tests/InterpolationTests.cs
+++ b/S2VX.Game.Tests/InterpolationTests.cs
@@ -1,6 +1,5 @@
 ﻿using NUnit.Framework;
 using osu.Framework.Graphics;
-using osu.Framework.Utils;
 using osuTK;
 using osuTK.Graphics;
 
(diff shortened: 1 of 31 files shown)


Xenocidel:
Is there a functional reason why we're changing the Alpha to be modified directly versus to a temp var first or is it just a style thing

maxrchung:
> Is there a functional reason why we're changing the Alpha to be modified directly versus to a temp var first or is it just a style thing

This way's a bit simpler. Saves a couple lines. We used to need a local variable because alpha was being applied separately to lines, release lines, and indicator lines for hold approaches. I changed it so the alpha just gets applied to the overall Approach. Doesn't seem to have any side effects from what I've tested on my map.

Xenocidel (approved):

Naranja-Sagged (approved):

PR #340, opened 2021-01-19
Grid optimization

### LINKED ISSUE #265: Optimize grid so that it doesn't recreate drawables every frame
As a developer, I want our grid to be more efficient so that it does not create a bunch of boxes every frame.

maxrchung:
Closes #265 

I looked into a few ways to try and reduce lag. Here are some optimizations in this PR:
* Remove any use of AlwaysPresent.
* Refactor grid logic so that new Drawables aren't created every update loop. New Drawables are added and removed only when needed.


Xenocidel (approved):
omg I can actually play the game now this is magic

Naranja-Sagged (approved):

PR #342, opened 2021-01-20
Command out of range handling

### LINKED ISSUE #339: Command out-of-range handling
As a story maker, I want the game to prevent me from adding commands with invalid (out of range) values, so that the game doesn't crash because it over/underflows to Infinity

maxrchung:
Closes #339 by using decorated float.Parse and double.Parse utility functions

### DIFF
--- a/S2VX.Game/Story/Command/S2VXCommand.cs
+++ b/S2VX.Game/Story/Command/S2VXCommand.cs
@@ -3,7 +3,6 @@ using Newtonsoft.Json.Linq;
 using osu.Framework.Graphics;
 using System;
 using System.Collections.Generic;
-using System.Globalization;
 using System.Linq;
 using System.Reflection;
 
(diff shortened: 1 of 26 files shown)


S2VX.Game/Story/Command/S2VXCommand.cs
```diff
@@ -56,8 +55,8 @@ public abstract class S2VXCommand : IComparable<S2VXCommand> {
             var systemType = Type.GetType($"S2VX.Game.Story.Command.{commandName}Command");
             var staticMethod = systemType.GetMethod("FromString", BindingFlags.Public | BindingFlags.Static);
             var command = staticMethod.Invoke(null, new object[] { split }) as S2VXCommand;
-            command.StartTime = double.Parse(split[1], CultureInfo.InvariantCulture);
-            command.EndTime = double.Parse(split[2], CultureInfo.InvariantCulture);
+            command.StartTime = S2VXUtils.StringToDouble(split[1]);
```
Xenocidel, line 58: I notice this is the only place that uses Doubles. All commands in practice use Floats.

Xenocidel (changes requested):

S2VX.Game/Story/Command/S2VXCommand.cs
maxrchung: This is intended to match how Drawables handle type values. Time values inside of Drawables are doubles. Position, rotation, scale, etc. are floats.

Xenocidel (approved):

Naranja-Sagged (approved):

PR #344, opened 2021-01-23
Story merge utility

### LINKED ISSUE #316: Story merge utility
As a developer, I want to be able to merge story files together so that I can combine our work together.

- Simple utility to take a couple story files and merge them together
- Check into repository
- Can be a ~~small script~~ or a C# utility. We decided to use C# since this may possibly be reused in the future (live editing xD).
- Probably want some kind of check for duplicates and output conflicts to be aware of

maxrchung:
Closes #316 

I decided to write this as a CLI utility. There are a number of ways you can use it. You can start a new Debug Instance of the StoryMerge project in Visual Studio and provide command line options through the StoryMerge project settings. You can build an executable, navigate to the executable, then run it directly.

Or, probably the most convenient, navigate to the StoryMerge project folder via command line and use dotnet run to execute the CLI:
```bash
cd StoryMerge
dotnet run -i input1.s2ry input2.s2ry -o output.s2ry
```

I probably went a little overboard with the test cases. The structure might make more sense if you view through Visual Studio/VSCode.


Xenocidel (changes requested):

StoryMerge/StoryMerger.cs
```diff
@@ -0,0 +1,217 @@
...
+            };
+            var message = string.Join("\n\n", messages);
+
+            return new Result {
+                IsSuccessful = true,
+                Message = message
+            };
+        }
+
+        public Result ValidateParameters() {
```
Xenocidel, line 47: Command line programs that I've used have always printed out the proper usage as if you typed in `--help` upon failure. Maybe we should add that here? Something like: `Usage: dotnet run -i input1.s2ry input2.s2ry -o output.s2ry`

Xenocidel (changes requested):

StoryMerge/CommandTimeInfo.cs
```diff
@@ -0,0 +1,27 @@
...
+
+        public override string ToString() =>
+            $"{Type} from {StartTime} to {EndTime}";
+
+        public int CompareTo(CommandTimeInfo other) {
+            var compare = StartTime.CompareTo(other.StartTime);
+            if (compare == 0) {
+                compare = EndTime.CompareTo(other.EndTime);
+            }
+            return compare;
```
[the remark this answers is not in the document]
maxrchung: I considered this, but it's not needed. CommandTimeInfo and NoteTimeInfo are classes I use to capture time info and compare times for conflicts. The types don't matter if it's only time conflicts that I care about.

StoryMerge/StoryMerger.cs
maxrchung: You can use `dotnet run -- --help` or `dotnet run -- -h` or `dotnet run -- -h` to get the help input. The help text and options are automatically provided by the command line library I'm using. The reason you need the extra -- in there is because dotnet run has its own CLI options. Adding the extra -- tells the dotnet run command to pass any further options into the command line program it's building and running.

More details: https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-run#options

StoryMerge/CommandTimeInfo.cs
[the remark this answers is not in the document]
Xenocidel: oh ok if it's only for conflicts then sure

StoryMerge/StoryMerger.cs
Xenocidel: I mean it seems inconsistent that it'd print out the usage if you have unrecognized/too many arguments but not when you have missing/too few arguments. It would be beneficial to the user to have the usage for the latter case. Not a huge deal, just a nitpick

maxrchung: Ah yeah, I get it now. I made a change to invoke the help command manually if merge was not successful.

Xenocidel (approved):

Naranja-Sagged (approved):

PR #345, opened 2021-01-23
Add EditorUI visibility container and refactor EditorScreen

### LINKED ISSUE #343: Hide all editor UI
As a historian, I would like to record a video of our finished map so that I can look back on it in the future with nostalgia.

* I think the simplest way to do this would be to hide all the UI panes in the EditorScreen then record the screen.
* We may have changes coming down the pipe that will make future S2RY maps incompatible with the existing ones. I don't think it's worth it to invest too much for backwards compatibility.

maxrchung:
Closes #343 

* Hide/show the editor UI by pressing the H key
* Rearrange some of the EditorScreen properties to maybe a better order
* Refactor some of the Load() function into smaller functions so it's better broken up

### DIFF
--- a/S2VX.Game/Editor/EditorScreen.cs
+++ b/S2VX.Game/Editor/EditorScreen.cs
@@ -23,14 +23,37 @@ using System;
 
 namespace S2VX.Game.Editor {
     public class EditorScreen : Screen {
-
         public const double TrackTimeTolerance = 0.03;  // ms away from a tick or end of track will be considered to be on that tick or end of track
+        private const int MaxSnapDivisor = 16;
 
+        public int SnapDivisor { get; private set; }
+        public Vector2 MousePosition { get; private set; } = Vector2.Zero;
+        public ReversibleStack Reversibles { get; } = new ReversibleStack();
+        public S2VXToolState ToolState { get; private set; } = new SelectToolState();
+
+        [Resolved]
+        private AudioManager Audio { get; set; }
         private string AudioPath { get; set; }
         private StorageBackedResourceStore CurLevelResourceStore { get; set; }
         private string CurSelectionPath { get; set; }
         private string StoryPath { get; set; }
         private string FullStoryPath { get; set; }
+
+        public DrawableTrack Track { get; private set; }
+        private S2VXStory Story { get; set; }
+        private EditorUI EditorUI { get; set; }
+        public Container<RelativeBox> NoteSelectionIndicators { get; } = new Container<RelativeBox> {
+            RelativePositionAxes = Axes.Both,
+            RelativeSizeAxes = Axes.Both,
+        };
+        private Container ToolContainer { get; set; } = new Container {
+            RelativeSizeAxes = Axes.Both,
+            Size = Vector2.One
+        };
+        public NotesTimeline NotesTimeline { get; } = new NotesTimeline();
+        private Timeline Timeline { get; } = new Timeline();
+        public CommandPanel CommandPanel { get; } = new CommandPanel();
+
         public EditorScreen(string curSelectionPath, string storyPath, StorageBackedResourceStore curLevelResourceStore, string audioPath) {
             CurSelectionPath = curSelectionPath;
             StoryPath = storyPath;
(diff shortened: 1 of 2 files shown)


S2VX.Game/Editor/EditorScreen.cs
```diff
@@ -98,100 +99,116 @@ public class EditorScreen : Screen {
...
-                                new MenuItem("Refresh (Ctrl+R)", ProjectRefresh),
-                                new MenuItem("Save (Ctrl+S)", ProjectSave),
-                                new MenuItem("Play (G)", ProjectPlay),
-                                new MenuItem("Quit (Esc)", ProjectQuit),
-                            }
-                        },
-                        new MenuItem("Edit")
+                            new MenuItem("Refresh (Ctrl+R)", ProjectRefresh),
+                            new MenuItem("Save (Ctrl+S)", ProjectSave),
+                            new MenuItem("Play (G)", ProjectPlay),
```
Xenocidel, line 142: Should we rename this Preview Gameplay to avoid confusion with Play/Pause?

S2VX.Game/Editor/EditorScreen.cs
```diff
@@ -98,100 +99,116 @@ public class EditorScreen : Screen {
...
+                        Items = new[]
                         {
-                            Items = new[]
-                            {
-                                new MenuItem("Command Panel (Ctrl+1)", ViewCommandPanel),
-                            }
-                        },
-                        new MenuItem("Playback")
+                            new MenuItem("Editor Interface (H)", ViewEditorUI),
+                            new MenuItem("Command Panel (Ctrl+1)", ViewCommandPanel),
```
Xenocidel, line 159: I kinda want to change this from Ctrl+1 to some single letter hotkey since it feels like an integral/essential part of story making that is used often.

S2VX.Game/Editor/EditorScreen.cs
```diff
@@ -98,100 +99,116 @@ public class EditorScreen : Screen {
...
-                                new MenuItem("Increase Playback Speed (Up,  MouseWheelUp over Speed)", PlaybackIncreaseRate),
-                                new MenuItem("Decrease Volume (MouseWheelDown over Volume)", VolumeDecrease),
-                                new MenuItem("Increase Volume (MouseWheelUp over Volume)", VolumeIncrease),
-                                new MenuItem("Decrease Snapping Divisor (MouseWheelDown over Snap Divisor)", SnapDivisorDecrease),
-                                new MenuItem("Increase Snapping Divisor (MouseWheelUp over Snap Divisor)", SnapDivisorIncrease),
-                            }
-                        },
-                        new MenuItem("Tool")
+                            new MenuItem("Play/Pause (Space)", PlaybackPlay),
+                            new MenuItem("Restart (X)", PlaybackRestart),
```
Xenocidel, line 167: Maybe we should make this not a single letter hotkey since it doesn't seem like it'd be used all that often. Maybe Ctrl+Left?

S2VX.Game/Editor/EditorScreen.cs
```diff
@@ -98,100 +99,116 @@ public class EditorScreen : Screen {
...
+                    {
+                        Items = new[]
                         {
-                            Items = new[]
-                            {
-                                new MenuItem("Command Panel (Ctrl+1)", ViewCommandPanel),
-                            }
-                        },
-                        new MenuItem("Playback")
+                            new MenuItem("Editor Interface (H)", ViewEditorUI),
```
Xenocidel, line 158: Not sure if I'd want this to be a single letter hotkey since
1. Probably not often used
2. If someone accidentally hits it they wouldn't know how to get back since everything is hidden

Xenocidel (changes requested):

S2VX.Game/Editor/EditorScreen.cs
maxrchung: Agreed

S2VX.Game/Editor/EditorScreen.cs
maxrchung: I will change it to just C

S2VX.Game/Editor/EditorScreen.cs
maxrchung: osu! editor uses X to go to beginning, and while I do use it a lot, it's probably more often on accident. Ctrl+Left makes sense.

S2VX.Game/Editor/EditorScreen.cs
maxrchung: I'll change to Ctrl+H.

Xenocidel (approved):

Naranja-Sagged (approved):

PR #361, opened 2021-02-09
Migrate code analyzers

### LINKED ISSUE #360: Migrate code analyzers
As a developer, I want to migrate our code analysis packages so that we are not using deprecated packages.

- Microsoft has migrated FxCopAnalyzers to NETAnalyzers, see this documentation: https://docs.microsoft.com/en-us/visualstudio/code-quality/migrate-from-fxcop-analyzers-to-net-analyzers?view=vs-2019

maxrchung:
Closes #360 

- Migrate deprecated FxCopAnalyzers to .NETAnalyzers
- Look through and reviewed our existing rules
- Refactor code according to new rules
- Update .NET version - you'll probably need to update Visual Studio and maybe install .NET 5 SDK
- Install dependabot and update packages


S2VX.Desktop/Program.cs
```diff
@@ -1,6 +1,10 @@
 ﻿using osu.Framework;
 using S2VX.Game;
+using System;
 
+// Yeah, so I'm not sure if we need to be compliant. I don't think it matters
+// unless we're aiming to target multiple .NET languages?
+[assembly: CLSCompliant(false)]
```
Xenocidel, line 7: What are we doing that is not compliant?

maxrchung: Ok, so this stuff's a bit confusing to me, but here's my understanding. There's the [Common Language Specification](https://docs.microsoft.com/en-us/dotnet/standard/language-independence-and-language-independent-components) that dictates rules that your code must conform to if you want that code to be able to run with different .NET languages. Since we aren't really exposing S2VX as a library or for other .NET languages besides C#, I don't think being CLS Compliant really matter to us.

But let's say we want to entertain the thought of being CLS Compliant. If you set `[assembly: CLSCompliant(true)]` in the and try to build, you'll find a lot of new CLS Compliant errors pop up, and that's actually because osu! framework itself is not marked as CLS Compliant. So now you'll have to go through everywhere that references osu! framework types and mark them with a `[CLSCompliant(false)]` attribute. At that point, I'd say that's too much busywork, and if the whole framework we're relying on isn't CLS Compliant, I don't think we can call ourselves compliant either.

Xenocidel: Ah ok, so it's osu! framework that's not compliant.

.editorconfig
```diff
@@ -239,16 +241,14 @@ dotnet_diagnostic.MEN007.severity = suggestion
 # MEN010: Avoid magic numbers
 dotnet_diagnostic.MEN010.severity = suggestion
 
-# MEN003: Method is too long
-dotnet_diagnostic.MEN003.severity = suggestion
```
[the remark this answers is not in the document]
Xenocidel: Oh ok, thought it was removing the enforcement. Read it wrong.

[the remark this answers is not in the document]
Xenocidel: I recall there's another rule somewhere that complains if our lines are longer than 160. I guess we'll just leave that be?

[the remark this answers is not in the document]
maxrchung: Yeah, the 160 line length is still enforced. We can configure it to be lower by adding some XML configuration files. I'm okay with 160 rn, but if you have a preference, we can look into modifying that.

S2VX.Game/Editor/ToolState/HoldNoteToolState.cs
```diff
@@ -30,7 +30,7 @@ public class HoldNoteToolState : S2VXToolState {
             Child = PreviewContainer;
         }
 
-        public override bool OnToolClick(ClickEvent _) {
+        public override bool OnToolClick(ClickEvent e) {
```
Naranja-Sagged, line 33: Are we not using the _ naming convention for unused variables anymore? I assume it's to follow CA1725: Parameter names should match base declaration

Naranja-Sagged (commented):
For some reason when I run S2VX on this branch it does not display my other stories, only the default Daboi story. On master I'm able to see and interact with my other stories.

S2VX.Game/Editor/ToolState/HoldNoteToolState.cs
maxrchung: Yep and yep

maxrchung:
> For some reason when I run S2VX on this branch it does not display my other stories, only the default Daboi story. On master I'm able to see and interact with my other stories.

Yeah, if you build the Desktop project and look at the resulting folder structure, it should make sense. Switching to .NET 5 creates a different output folder that's specific for .NET 5. There should be an existing folder that we previously used for .NET 3.1. You can copy over the Stories from .NET 3.1 to the .NET 5 folder.

Naranja-Sagged:
> > For some reason when I run S2VX on this branch it does not display my other stories, only the default Daboi story. On master I'm able to see and interact with my other stories.
> 
> Yeah, if you build the Desktop project and look at the resulting folder structure, it should make sense. Switching to .NET 5 creates a different output folder that's specific for .NET 5. There should be an existing folder that we previously used for .NET 3.1. You can copy over the Stories from .NET 3.1 to the .NET 5 folder.

Oh ok I see them now pogs

Naranja-Sagged (approved):

Xenocidel (approved):

PR #385, opened 2021-02-25
Disllow clicks outside MissThreshold for GameNotes

### LINKED ISSUE #383: Clickable outside MissThreshold
As an S2VX player, I do not want to be able to hit notes outside of miss threshold as my fingers can slip sometimes so I do not want to be overly punished with a miss.

Xenocidel:
Closes #383

### DIFF
--- a/S2VX.Game/Story/Note/GameNote.cs
+++ b/S2VX.Game/Story/Note/GameNote.cs
@@ -20,8 +20,8 @@ namespace S2VX.Game.Story.Note {
 
         private Sample Hit { get; set; }
         private Sample Miss { get; set; }
-        public const int MissThreshold = 200;
-        private int TimingError;
+        public const double MissThreshold = 200.0;
+        private double TimingError;
         private bool IsFlaggedForRemoval { get; set; }
         public override bool HandlePositionalInput => true;
 
@@ -32,7 +32,7 @@ namespace S2VX.Game.Story.Note {
         }
 
         private void FlagForRemoval() {
-            PlayScreen.PlayInfoBar.RecordHitError(TimingError);
+            PlayScreen.PlayInfoBar.RecordHitError((int)TimingError);
             if (Math.Abs(TimingError) < MissThreshold) {
                 Hit.Play();
             } else {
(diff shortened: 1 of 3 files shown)


Xenocidel:
I moved `Story.Notes.HasClickedNote = true;` down to `ClickNote()` because I think it makes more sense for that flag to be applied there rather than a function meant for checking if a note is clickable

S2VX.Game/Story/Note/GameNote.cs
```diff
@@ -53,17 +53,19 @@ public class GameNote : S2VXNote, IKeyBindingHandler<InputAction> {
...
-            // Limit timing error to be +/- MissThreshold (though it will never be >= MissThreshold since RecordMiss would have already run)
-            TimingError = (int)Math.Round(Math.Clamp(time - HitTime, -MissThreshold, MissThreshold));
             var isVisible = Alpha > 0;
-            Story.Notes.HasClickedNote = true;
-            return isVisible;
+            var isWithinMissThreshold = Math.Abs(time - HitTime) <= MissThreshold;
+            if (!isVisible || !isWithinMissThreshold) {
+                return false;
+            }
+            TimingError = (int)Math.Round(time - HitTime);
```
maxrchung, line 62:
```suggestion
            TimingError = time - HitTime;
```

One change I made was to make score a `double` for accuracy. The `ScoreInfo` class handles rounding to an `int`. Can we remove the rounding stuff here and see what happens?

Xenocidel: Should I change MissThreshold to be a double as well?

maxrchung: Yeah, I think that would make sense and calculations easier.

Xenocidel: Alright I'll change it for GameNote and GameHoldNote but still pass in int to the ScoreBar since we're gonna get rid of it later anyways and I don't want to refactor it in this change.

maxrchung (approved):

Naranja-Sagged (approved):

PR #394, opened 2021-03-06
Change hold note scoring

### LINKED ISSUE #391: Change HoldNote scoring
As a player, I don't want my score to be calculated outside of MissThreshold so that my score is calculated more as expected.

Xenocidel:
Closes #391

### DIFF
--- a/S2VX.Game.Tests/VisualTests/GameHoldNoteTests.cs
+++ b/S2VX.Game.Tests/VisualTests/GameHoldNoteTests.cs
@@ -30,9 +30,15 @@ namespace S2VX.Game.Tests.VisualTests {
         [SetUpSteps]
         public void SetUpSteps() {
             AddStep("Clear score", () => PlayScreen.ScoreInfo.ClearScore());
-            AddStep("Reset story", () =>
-            Story.Reset());
+            AddStep("Reset story", () => Story.Reset());
             AddStep("Reset clock", () => Story.Clock = new FramedClock(Stopwatch = new StopwatchClock()));
+        }
+
+        public void PressAndRelease(double timeToPress, double holdDuration) {
+            AddStep($"Seek to {timeToPress}", () => Stopwatch.Seek(timeToPress));
+            AddStep("Press note", () => InputManager.PressKey(Key.Z));
+            AddStep($"Seek to {timeToPress + holdDuration}", () => Stopwatch.Seek(timeToPress + holdDuration));
+            AddStep("Release note", () => InputManager.ReleaseKey(Key.Z));
         }
 
         // Assuming defaults are as such:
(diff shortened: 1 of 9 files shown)


Xenocidel:
Please refer to https://github.com/maxrchung/S2VX/wiki/Score-Calculation for a visual diagram of the scoring structure.

Xenocidel:
This change also changes Story's `AddHoldNote` to an overloaded `AddNote` to avoid mishaps with calling the incorrect function when attempting to add hold notes.

S2VX.Game.Tests/VisualTests/GameHoldNoteTests.cs
```diff
@@ -52,8 +125,137 @@ public class GameHoldNoteTests : S2VXTestScene {
             AddUntilStep("Wait until all notes are removed", () => Story.Notes.Children.Count == 0);
             AddStep("Release key", () => InputManager.ReleaseKey(Key.Z));
             AddAssert("Does not trigger multiple notes", () =>
-                Precision.AlmostEquals(originalNoteCount * 100, PlayScreen.ScoreInfo.Score, 20)
+                Precision.AlmostEquals((originalNoteCount - 1) * 100, PlayScreen.ScoreInfo.Score, 20)
```
maxrchung, line 128: Since score is more exact now, can we remove using `AlmostEquals`?

Xenocidel: Done

maxrchung (approved):

S2VX.Game.Tests/VisualTests/GameHoldNoteTests.cs
```diff
@@ -51,9 +124,136 @@ public class GameHoldNoteTests : S2VXTestScene {
             AddStep("Start clock", () => Stopwatch.Start());
             AddUntilStep("Wait until all notes are removed", () => Story.Notes.Children.Count == 0);
             AddStep("Release key", () => InputManager.ReleaseKey(Key.Z));
-            AddAssert("Does not trigger multiple notes", () =>
-                Precision.AlmostEquals(originalNoteCount * 100, PlayScreen.ScoreInfo.Score, 20)
-            );
+            AddAssert("Does not trigger multiple notes", () => PlayScreen.ScoreInfo.Score == (originalNoteCount - 1) * 100);
+        }
+
+        [TestCase(20, 20)]  // Press and Release entirely within NotVisible state
```
Naranja-Sagged, line 130: Lol didn't know you could do this. This is nuts.

S2VX.Game/Story/Note/GameHoldNote.cs
```diff
@@ -19,25 +19,37 @@ public class GameHoldNote : HoldNote, IKeyBindingHandler<InputAction> {
         [Resolved]
         private PlayScreen PlayScreen { get; set; }
 
-        private enum HoldNoteState {
-            NotVisibleBefore,
+        public enum HoldNoteState {
```
Naranja-Sagged, line 22: Wasn't it good practice to separate enums into their own files? Might need @maxrchung 's opinion on this.

maxrchung: Yeah, I agree. I think we made an exception for private enums since those would never be used outside of that file, but public enums or classes should be put into a separate file.

Xenocidel: The test is the only place that uses HoldNoteState outside of this file. Otherwise it could be private. How should I proceed? Still put it in a separate file?

maxrchung: Yeah, my preference is for all public enums and classes to be in a separate file.

maxrchung: For more context, this is the rule that suggests not having nested public types: https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1034

I disabled this rule because NUnit tests can sometimes be written as one class per test, see [StoryMerge.Tests](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1034) that have examples of this.

Naranja-Sagged (approved):

PR #406, opened 2021-03-15
Cursor color rotation scale

### LINKED ISSUE #126: Cursor color, rotation, scale

maxrchung:
Closes #126


maxrchung:
Looking into issues with Cursor dependencies

PR #413, opened 2021-03-23
Fix visibility of song preview tests

### LINKED ISSUE #404: Fix visibility of SongPreviewTests
As a developer I want the elements within the SongPreviewTests Visual Test to be visible so that I can verify with my eyes in the GUI that it looks correct

maxrchung:
Closes #404 

* Changed `SongPreviewTests` to `[Headless]` since they visually don't make sense
* Refactored `SongPreview` that it uses the `S2VXGameBase.GameWidth`
* Moved tests within a folder for what I think is better organization
* Added some more comments to note some differences between `S2VXTestBrowser` and `S2VXTestSceneTestRunner`


S2VX.Game.Tests/S2VXTestBrowser.cs
```diff
@@ -1,17 +1,30 @@
...
+    /// class.
+    /// </summary>
     public class S2VXTestBrowser : S2VXGameBase {
+        /// <summary>
+        /// S2VXStory resolves an S2VXCursor as one of its dependencies. To
+        /// avoid having to manually cache a Cursor into many visual tests, we
+        /// can cache it here.
+        /// </summary>
+        [Cached]
+        private S2VXCursor Cursor { get; set; } = new();
```
Xenocidel, line 19: We can move this to S2VXGameBase so we don't need to create and cache it here AND in S2VXGame

Xenocidel (changes requested):

S2VX.Game.Tests/S2VXTestBrowser.cs
maxrchung: I remember testing this yesterday without great results because of the way S2VXGameBase sets its Content. I'll try a few thing more tonight to confirm.

maxrchung: From what I tested, I plan to keep the changes as is. I tried moving `Cursor` into `GameBase`, but ran into issues because of the way the container's content is being set. I also tried moving it into `SquareContainer`, but ran into problems with the load order of the dependencies.

Xenocidel: Moved to S2VXGameBase

Xenocidel (approved):

S2VX.Game/SongSelection/Containers/SongPreview.cs
```diff
@@ -121,13 +120,11 @@ public class SongPreview : CompositeDrawable {
                     }
                 },
                 new FillFlowContainer {
```
Xenocidel, line 122: This container is no longer centered. Looks like it got larger too

Xenocidel (changes requested):

Xenocidel (approved):

Naranja-Sagged (approved):

PR #416, opened 2021-03-29
Score processor for regular notes

### LINKED ISSUE #401: ScoreProcessor for regular notes
As a developer, I want to separate score logic into its own class so that we can reuse it for all notes and decouple score logic from notes.

image

New system:
image

Determine how the hit window divisions will work

maxrchung:
Closes #401


S2VX.Game/Play/UserInterface/ScoreProcessor.cs
```diff
@@ -0,0 +1,87 @@
...
+
+        public void AddScore(double moreScore) {
+            Score += moreScore;
+            TxtScore.Text = $"{Math.Round(Score)}";
+        }
+
+        public void Reset() {
+            Score = 0;
+            TxtScore.Text = $"{Math.Round(Score)}";
+            Hit.Reset();
```
Xenocidel, line 57: Does resetting the samples actually do anything?

S2VX.Game/Play/UserInterface/ScoreProcessor.cs
```diff
@@ -0,0 +1,87 @@
...
+            Miss.Reset();
+        }
+
+        public void Process(double time, GameNote note) {
+            var notes = Story.Notes;
+            var score = Math.Abs(time - note.HitTime);
+
+            if (score > notes.MissThreshold) { // miss
+                AddScore(notes.MissThreshold);
```
Xenocidel, line 66: I was wondering if we'll ever need to know the state/result and thus would it be nice to have an enum for that (miss, early, late, etc.). But I guess if we're not using it now there's no need in adding in dead code. Just a thought for the future.

S2VX.Game/Play/UserInterface/ScoreProcessor.cs
Xenocidel: Oh never mind I see it resets the PlayCount variable

S2VX.Game/Story/Note/Notes.cs
```diff
@@ -27,6 +27,14 @@ public class Notes : CompositeDrawable {
         public Color4 PreviewNoteColor { get; set; }
         public Color4 PreviewHoldNoteColor { get; set; }
 
+        // TODO: Remove these defaults and use command classes
+        public Color4 PerfectColor { get; set; } = Color4.Green;
```
Xenocidel, line 31: Should we use S2VXColorConstants rather than Color4?

Xenocidel:
After seeing this, I think the cursor changing color is probably not a very good indicator of miss/early/late/perfect. The notes come way too fast so the color changes before you can see it. Probably better to just leave the cursor's color static and instead do osu!'s method of having the error indicators persist where the note was.

S2VX.Game.Tests/HeadlessTests/ScoreProcessorTests.cs
```diff
@@ -0,0 +1,88 @@
...
+using osu.Framework.Audio;
+using osu.Framework.Screens;
+using osu.Framework.Testing;
+using S2VX.Game.Play;
+using S2VX.Game.Play.UserInterface;
+using S2VX.Game.Story;
+using S2VX.Game.Story.Note;
+using System.IO;
+
+namespace S2VX.Game.Tests.VisualTests {
```
Naranja-Sagged, line 12: Is it supposed to be S2VX.Game.Tests.VisualTests here? I can't find these tests under the testing UI or test explorer.

Naranja-Sagged:
Is it possible to trigger an early miss right now? My early keypresses seem to all be ignored and never triggers early misses.

Naranja-Sagged:
> After seeing this, I think the cursor changing color is probably not a very good indicator of miss/early/late/perfect. The notes come way too fast so the color changes before you can see it. Probably better to just leave the cursor's color static and instead do osu!'s method of having the error indicators persist where the note was.

I agree the cursor flashing is hard to follow so I find it hard to tell if I'm hitting early or late. I think it's useful for when cursor is mostly static and player is hitting stacked streams but for jumps it isn't too useful

Xenocidel:
> Is it possible to trigger an early miss right now? My early keypresses seem to all be ignored and never triggers early misses.

Addressed in new patch

S2VX.Game.Tests/HeadlessTests/ScoreProcessorTests.cs
Xenocidel: Addressed in new patch

Xenocidel (approved):

Naranja-Sagged (approved):

PR #444, opened 2021-04-16
Apply command to note properties and sync Hold Path Color with Outline Color

### LINKED ISSUE #264: Commands affect Notes properties instead of Note directly
As a developer, I want to have more efficient code to apply commands so that only one property is updated by a command.

We have a few commands that are modifying a Note directly. It would be better separation if we modified a higher level property inside of Notes and then referred to this property in a Note's Update.

- Make sure to apply color to Slider path and Anchor path

### LINKED ISSUE #270: Add outline to Preview notes and Preview hold notes
As an editor, I want the preview notes to have an outline so that it looks visually the same as a placed note.

Apparently this was caused by #257

### LINKED ISSUE #277: HoldNote's Color is not set on initial creation
As an editor, I want the hold note to have its proper color when it is first created so that it is consistent with how it's supposed to appear.

- Initial create doesn't have the right color
- After you move the time i.e. with the scroll wheel, the color is set properly

### LINKED ISSUE #279: Add back(?) preview for hold note tool
As an editor, I want there to be preview indicators when placing hold notes so that I can accurately place hold notes.

- Currently there's no preview note that shows up at all
- If possible, it'd be nice to also see the slider being drawn in real time

Xenocidel:
Closes #264 
Also happens to close #270
Also happens to close #277 
Also happens to close #279

### DIFF
--- a/S2VX.Game.Tests/HeadlessTests/EditorHoldNoteTests.cs
+++ b/S2VX.Game.Tests/HeadlessTests/EditorHoldNoteTests.cs
@@ -5,6 +5,7 @@ using osu.Framework.Screens;
 using osu.Framework.Testing;
 using S2VX.Game.Editor;
 using S2VX.Game.Story;
+using S2VX.Game.Story.Command;
 using S2VX.Game.Story.Note;
 using System.IO;
 
@@ -37,6 +38,10 @@ namespace S2VX.Game.Tests.HeadlessTests {
             AddStep("Add note", () => Story.AddNote(NoteToTest = new EditorHoldNote {
                 HitTime = Story.Notes.ShowTime + Story.Notes.FadeInTime + NoteAppearTime,
                 EndTime = Story.Notes.ShowTime + Story.Notes.FadeInTime + NoteAppearTime + HoldDuration
+            }));
+            AddStep("Set max note alpha to 1", () => Story.AddCommand(new HoldNotesAlphaCommand {
+                StartValue = 1.0f,
+                EndValue = 1.0f
             }));
         }
 
--- a/S2VX.Game/Story/Command/HoldNotesColorCommand.cs
+++ b/S2VX.Game/Story/Command/HoldNotesColorCommand.cs
@@ -6,8 +6,8 @@ namespace S2VX.Game.Story.Command {
         public Color4 EndValue { get; set; } = S2VXColorConstants.Cyan;
         public override void Apply(double time, S2VXStory story) {
             var value = S2VXUtils.ClampedInterpolation(time, StartValue, EndValue, StartTime, EndTime, Easing);
-            story.Notes.GetHoldNotes().ForEach(note => note.SetColor(value));
             story.Notes.PreviewHoldNoteColor = value;
+            story.Notes.HoldNoteColor = value;
         }
         protected override string ToValues() => $"{S2VXUtils.Color4ToString(StartValue)}|{S2VXUtils.Color4ToString(EndValue)}";
         public static HoldNotesColorCommand FromString(string[] split) {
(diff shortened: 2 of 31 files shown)


S2VX.Game.Tests/HeadlessTests/EditorHoldNoteTests.cs
```diff
@@ -38,6 +39,10 @@ public class EditorHoldNoteTests : S2VXTestScene {
                 HitTime = Story.Notes.ShowTime + Story.Notes.FadeInTime + NoteAppearTime,
                 EndTime = Story.Notes.ShowTime + Story.Notes.FadeInTime + NoteAppearTime + HoldDuration
             }));
+            AddStep("Set max note alpha to 1", () => Story.AddCommand(new HoldNotesAlphaCommand {
+                StartValue = 1.0f,
+                EndValue = 1.0f
+            }));
```
maxrchung, line 45: Why do we need these?

S2VX.Game/Story/Command/HoldNotesColorCommand.cs
```diff
@@ -6,8 +6,8 @@ public class HoldNotesColorCommand : S2VXCommand {
         public Color4 EndValue { get; set; } = S2VXColorConstants.Cyan;
         public override void Apply(double time, S2VXStory story) {
             var value = S2VXUtils.ClampedInterpolation(time, StartValue, EndValue, StartTime, EndTime, Easing);
-            story.Notes.GetHoldNotes().ForEach(note => note.SetColor(value));
             story.Notes.PreviewHoldNoteColor = value;
```
maxrchung, line 9: I think we should just have `HoldNoteColor` and have the preview hold note pull from that value. Same with `PreviewNotecolor`.

S2VX.Game/Story/Note/S2VXNote.cs
```diff
@@ -11,12 +11,14 @@ public abstract class S2VXNote : CompositeDrawable, IComparable<S2VXNote> {
...
         public Vector2 Coordinates { get; set; } = Vector2.Zero;
         public virtual Approach Approach { get; set; }
-        public float OutlineThickness { get; set; }
-        public Color4 OutlineColor { get; set; }
 
         protected RelativeBox BoxOuter { get; } = new RelativeBox();
         protected RelativeBox BoxInner { get; } = new RelativeBox();
 
+        public Color4 InnerColor { get => BoxInner.Colour; set => BoxInner.Colour = value; }
+        public Color4 OutlineColor { get => BoxOuter.Colour; set => BoxOuter.Colour = value; }
```
maxrchung, line 19: Don't know if this indirection is really necessary. I think I'd rather just make `BoxInner` and `BoxOuter` public.

S2VX.Game.Tests/HeadlessTests/EditorHoldNoteTests.cs
Xenocidel: The default alpha for HoldNotes has been changed to 0.7 in HoldNotesAlphaCommand.cs
Explicitly setting the alpha to 1.0 in SetUpSteps is now needed to keep the test expectations consistent. I figured it was better than hard coding all the expectations to 0.7 since we may change that default again.

S2VX.Game/Story/Command/HoldNotesColorCommand.cs
Xenocidel: Implemented. I'll have the Preview note pull from the same constants for InnerColor, OutlineColor, and OutlineThickness. I'll leave Alpha alone so you can easily differentiate it from the notes already placed.

S2VX.Game/Story/Note/S2VXNote.cs
Xenocidel: These used to be dedicated Getter/Setter functions but Intellectus recommended that I change it to this format. I prefer to keep them like this because later on if we decide to change this (like in #262 ), we wouldn't have to change and remember the names of the objects within the S2VXNote whenever we want to do something with the note color.

maxrchung (approved):

Naranja-Sagged (approved):

PR #454, opened 2021-04-26
Separate ar for editor

### LINKED ISSUE #323: Would be nice to be able to see more notes in editor (separate AR for editor?)
As an editor, I want to be able to slow down the AR so that I can see where the note patterns are.

- Possible solution is to add another scrolling input in the editor to change the AR. Could set something like the snap divisor with 2x, 4x, 6x, 8x maybe possibly idk - up to implementer

Naranja-Sagged:
Closes #323


Xenocidel:
Tests are failing

Naranja-Sagged:
> Tests are failing

Oops lol, I think changing how showtime works broke a few tests

Issue #459, opened 2021-05-02
Fix NotesTimeline visual tests (GUI)


Xenocidel:
As a developer, I want to fix the GUI for visual tests so that coordinates/offsets relative to object positions are accurate.

For some resolutions, the NotesTimeline visual tests in the GUI will fail due to the scaling of the DrawSizePreservingFillContainer within the test runner scene.

maxrchung:
I looked at this earlier in the week but didn't make much more progress other than what Aaron already looked at. I think that to troubleshoot this further, one thing we can do for ourselves is to setup our S2VX and osu-framework solutions in such a way that we can hook up osu-framework to our running code.

PR #460, opened 2021-05-02
Notes timeline Scrolling

### LINKED ISSUE #455: Scroll note timeline when timeline note is dragged to edges
As an editor, I want the note timeline to scroll when a timeline note is dragged to the edges so that I don't have to use left or right to separately move the time.

Xenocidel:
Closes #455 
Causes #459

### DIFF
--- a/S2VX.Game/Editor/Containers/NotesTimeline.cs
+++ b/S2VX.Game/Editor/Containers/NotesTimeline.cs
@@ -20,6 +20,17 @@ namespace S2VX.Game.Editor.Containers {
         [Resolved]
         private S2VXStory Story { get; set; }
 
+        // Tick information is calculated by GetTickInfo for a given time
+        private struct TickInfo {
+            public double NumTicks;
+            public double TimeBetweenTicks;
+            public double CurrentTick;
+            public double CurrentTickTime;
+            public int NearestTick;
+            public double NearestTickTime;
+            public bool OnTick;
+        }
+
         public Dictionary<S2VXNote, double> SelectedNoteToTime { get; private set; } = new Dictionary<S2VXNote, double>();
         public Dictionary<S2VXNote, RelativeBox> NoteToTimelineNote { get; } = new Dictionary<S2VXNote, RelativeBox>();
 
@@ -109,7 +120,7 @@ namespace S2VX.Game.Editor.Containers {
             new() {
                 RelativePositionAxes = Axes.Both,
                 RelativeSizeAxes = Axes.Both,
-                Width = TimelineWidth / 1.25f,
+                Width = TimelineWidth * 0.8f,
                 Anchor = Anchor.Centre,
                 Origin = Anchor.Centre,
                 X = -0.05f,
@@ -229,23 +240,38 @@ namespace S2VX.Game.Editor.Containers {
                 }
             };
 
-        public void SnapToTick(bool snapLeft) {
+        public double GetNearestTickTime(double time) => GetTickInfo(time).NearestTickTime;
+
+        private TickInfo GetTickInfo(double time) {
             var numTicks = Story.BPM / 60f * (Editor.Track.Length / SecondsToMS) * Divisor;
             var timeBetweenTicks = Editor.Track.Length / numTicks;
-            var currentTick = (Time.Current - Story.Offset) / timeBetweenTicks;
+            var currentTick = (time - Story.Offset) / timeBetweenTicks;
             var currentTickTime = currentTick * timeBetweenTicks;
-            var nearestTick = Math.Round(currentTick);
+            var nearestTick = (int)Math.Round(currentTick);
             var nearestTickTime = nearestTick * timeBetweenTicks;
             var onTick = Math.Abs(nearestTickTime - currentTickTime) <= EditorScreen.TrackTimeTolerance;
-            var leftTickTime = timeBetweenTicks;
-            var rightTickTime = timeBetweenTicks;
-            if (onTick) {
-                leftTickTime *= nearestTick - 1;
-                rightTickTime *= nearestTick + 1;
+            return new TickInfo {
+                NumTicks = numTicks,
+                TimeBetweenTicks = timeBetweenTicks,
+                CurrentTick = currentTick,
+                CurrentTickTime = currentTickTime,
+                NearestTick = nearestTick,
+                NearestTickTime = nearestTickTime,
+                OnTick = onTick
+            };
+        }
+
+        public void SnapToTick(bool snapLeft) {
+            var tickInfo = GetTickInfo(Time.Current);
+            var leftTickTime = tickInfo.TimeBetweenTicks;
+            var rightTickTime = tickInfo.TimeBetweenTicks;
+            if (tickInfo.OnTick) {
+                leftTickTime *= tickInfo.NearestTick - 1;
+                rightTickTime *= tickInfo.NearestTick + 1;
             } else {
                 // Current time is between two ticks
-                leftTickTime *= Math.Floor(currentTick);
-                rightTickTime *= Math.Ceiling(currentTick);
+                leftTickTime *= Math.Floor(tickInfo.CurrentTick);
+                rightTickTime *= Math.Ceiling(tickInfo.CurrentTick);
             }
 
             leftTickTime += Story.Offset;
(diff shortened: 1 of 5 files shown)


S2VX.Game/Editor/Containers/NotesTimeline.cs
```diff
@@ -20,10 +20,25 @@ public class NotesTimeline : CompositeDrawable {
...
+        }
+
+        public double NumTicks { get; private set; }
+        public double TimeBetweenTicks { get; private set; }
+        private int NumVisibleTicks { get; set; }
+        public int FirstVisibleTick { get; private set; }
+        public int LastVisibleTick { get; private set; }
+
         public Dictionary<S2VXNote, double> SelectedNoteToTime { get; private set; } = new Dictionary<S2VXNote, double>();
         public Dictionary<S2VXNote, RelativeBox> NoteToTimelineNote { get; } = new Dictionary<S2VXNote, RelativeBox>();
```
maxrchung, line 39: Can do some cleanup with `new()`

S2VX.Game/Editor/Containers/NotesTimeline.cs
```diff
@@ -20,10 +20,25 @@ public class NotesTimeline : CompositeDrawable {
...
+            public int NearestTick;
+            public double NearestTickTime;
+            public bool OnTick;
+        }
+
+        public double NumTicks { get; private set; }
+        public double TimeBetweenTicks { get; private set; }
+        private int NumVisibleTicks { get; set; }
+        public int FirstVisibleTick { get; private set; }
+        public int LastVisibleTick { get; private set; }
```
maxrchung, line 36: I think some of these properties can be private

S2VX.Game/Editor/ToolState/SelectToolState.cs
```diff
@@ -192,41 +217,42 @@ public class SelectToolState : S2VXToolState {
...
-            mousePosX = Math.Clamp(mousePosX, 0, noteSelectionIndicators.DrawWidth);
-            var relativeMousePosX = mousePosX / noteSelectionIndicators.DrawWidth;
-            var gameTimeDeltaFromMiddle = (relativeMousePosX - 0.5f) * Editor.NotesTimeline.SectionLength * NotesTimeline.SecondsToMS;
-            return Time.Current + gameTimeDeltaFromMiddle;
-        }
-
         public override void OnToolDrag(DragEvent e) {
             if (!DelayDrag) {
                 switch (ToDrag) {
                     case SelectToolDragState.DragHoldNoteEndTime: {
```
maxrchung, line 223: Since this file's getting pretty big with nested switch cases, a refactoring suggestion would be to make `DragState` its own state class that implements `DragStart`, `Drag`, `DragEnd` functions.

Naranja-Sagged:
s2vx dragnote
Some really weird behavior when notes next to edge of notes timeline are dragged either left or right.
In this image if I drag the 3rd pink note starting from the left it disappears. If I drag the 1st pink note to the right it scrolls the notes timeline to the left a bit and stops after I drag it a distance. This is more noticeable when zoomed in closer to the notes timeline like in my image

Naranja-Sagged:
s2vx scroll offset
Sometimes dragging to the right with a note can mess up the offset from cursor to note in the notes timeline. In the image I scrolled a bit to the right with the selected note and am still dragging. Dragging any of the other notes on that timeline would also have the same offset problem

S2VX.Game/Editor/ToolState/SelectToolState.cs
Xenocidel: So this would be another ToolState that we'd switch to when Dragging?

S2VX.Game/Editor/Containers/NotesTimeline.cs
Xenocidel: Implemented

S2VX.Game/Editor/Containers/NotesTimeline.cs
Xenocidel: Cleaned up some unneeded variables

Xenocidel:
> s2vx scroll offset
> Sometimes dragging to the right with a note can mess up the offset from cursor to note in the notes timeline. In the image I scrolled a bit to the right with the selected note and am still dragging. Dragging any of the other notes on that timeline would also have the same offset problem

I fixed this bug (has to do with stories with offset) in the latest commit. Could you check to see if it's resolved?

Naranja-Sagged:
> > s2vx scroll offset
> > Sometimes dragging to the right with a note can mess up the offset from cursor to note in the notes timeline. In the image I scrolled a bit to the right with the selected note and am still dragging. Dragging any of the other notes on that timeline would also have the same offset problem
> 
> I fixed this bug (has to do with stories with offset) in the latest commit. Could you check to see if it's resolved?

Yea this is fixed now 👌

Xenocidel:
> s2vx dragnote
> Some really weird behavior when notes next to edge of notes timeline are dragged either left or right.
> In this image if I drag the 3rd pink note starting from the left it disappears. If I drag the 1st pink note to the right it scrolls the notes timeline to the left a bit and stops after I drag it a distance. This is more noticeable when zoomed in closer to the notes timeline like in my image

I can't seem to reproduce the note disappearing issue. Maybe that's fixed with the latest commit?
Regarding the second scenario you mentioned, yes that occurs because I only look at whether or not your mouse position is nearest the first/last tick. I can improve it by adding logic to take into account cursor direction. Will address in next commit.

Naranja-Sagged:
> > s2vx dragnote
> > Some really weird behavior when notes next to edge of notes timeline are dragged either left or right.
> > In this image if I drag the 3rd pink note starting from the left it disappears. If I drag the 1st pink note to the right it scrolls the notes timeline to the left a bit and stops after I drag it a distance. This is more noticeable when zoomed in closer to the notes timeline like in my image
> 
> I can't seem to reproduce the note disappearing issue. Maybe that's fixed with the latest commit?
> Regarding the second scenario you mentioned, yes that occurs because I only look at whether or not your mouse position is nearest the first/last tick. I can improve it by adding logic to take into account cursor direction. Will address in next commit.

Yea it seems like the disappearing bug was fixed by the last commit

S2VX.Game/Editor/ToolState/SelectToolState.cs
maxrchung: No, it would be implemented similarly to `ToolState`, but it's only maintained in the `SelectToolState` as a substate, if that makes sense. For instance, some pseudocode:

```csharp
interface ISelectToolDragState {
    OnToolDragStart();
    OnToolDrag();
    OnToolDragEnd();
}

class DragTimelineNote : ISelectToolDragState {
    DragTimelineNote(SelectToolState toolState) { ... }
    OnToolDragStart() { ... }
    OnToolDrag() { ... }
    OnToolDragEnd() { ... }
}

class AnotherDragState : ISelectToolDragState {
    DragTimelineNote(SelectToolState toolState) { ... }
    OnToolDragStart() { ... }
    OnToolDrag() { ... }
    OnToolDragEnd() { ... }
}

class SelectToolState {
    ISelectToolDragState DragState;

    OnToolDragStart() {
        if (someCondition) {
            DragState = new DragTimelineNote(this);
        } else if (someOtherCondition) {
            DragState = new AnotherDragState(this);
        } else {
            ...
        }
        DragState.DragStart();
    }

    OnToolDrag() {
        DragState.Drag();
    }

    OnToolDragEnd() {
        DragState.DragEnd();
    }
}
```

Naranja-Sagged:
The scrolling is harder to control the more zoomed in the notes timeline is. It's really hard to be precise with note time placement with really close zooms. Osu! stable uses some kind of system where the scroll distance scales with how zoomed out it is, I think that would make the scrolling easier to use

Xenocidel:
> The scrolling is harder to control the more zoomed in the notes timeline is. It's really hard to be precise with note time placement with really close zooms. Osu! stable uses some kind of system where the scroll distance scales with how zoomed out it is, I think that would make the scrolling easier to use

This is a good point. This happens because the scrolling is done every time OnToolDrag is called. So if the mouse moves more (which is the case when the timeline is more zoomed in), then it will scroll more. I'll have to think of how to change this to make it easier to control.

maxrchung (approved):

Naranja-Sagged (approved):

PR #461, opened 2021-05-03
Implement color input in command panel

### LINKED ISSUE #448: Implement color input in command panel
As a developer, I want to implement the color input into the command panel so that we can use the color input to easily select colors.

maxrchung:
Closes #448

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,31 @@
+version: 2
+updates:
+- package-ecosystem: nuget
+  directory: "/"
+  schedule:
+    interval: daily
+    time: "13:00"
+  open-pull-requests-limit: 10
+  ignore:
+  - dependency-name: ppy.osu.Framework
+    versions:
+    - 2021.204.0
+    - 2021.413.0
+    - 2021.415.0
+    - 2021.422.1
+    - 2021.424.1
+  - dependency-name: Menees.Analyzers
+    versions:
+    - 2.0.6
+  - dependency-name: NUnit3TestAdapter
+    versions:
+    - 3.17.0
+  - dependency-name: Microsoft.NET.Test.Sdk
+    versions:
+    - 16.8.3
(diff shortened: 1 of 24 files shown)


Xenocidel:
The first time you click to open the ColorPicker it always defaults to (0,0,0) rather than applying the color that was already in the field.

maxrchung:
> The first time you click to open the ColorPicker it always defaults to (0,0,0) rather than applying the color that was already in the field. Subsequent clicks to open do not revert back to black.
> EDIT: Looks like if you copy another command up and open the ColorPicker it jumps to a different color altogether.

Should be fixed with https://github.com/maxrchung/S2VX/pull/461/commits/12b9d6db63144ae36ec571fee4d8f8d9fff8214d. The code comments mention a possible recursion issue, but I did not encounter it.

Naranja-Sagged:
I think only 1 color picker should be able to be open at the same time since having 2 open looks kinda jank. The color picker on the right always overlaps the one on the left so I don't think it's too useful

maxrchung:
> I think only 1 color picker should be able to be open at the same time since having 2 open looks kinda jank. The color picker on the right always overlaps the one on the left so I don't think it's too useful

I considered this, but after prototyping and testing a bit, I think it was kind of hard to hook up the value inputs to a single color picker. So, at least coding design-wise, I think this is easier to reason with. As for the overlap issues, I can rearrange the inputs and adjust the input sizing.

image

Xenocidel:
When editing a command, the color picker is under the other commands
image

maxrchung:
> When editing a command, the color picker is under the other commands
> image

Fixed in https://github.com/maxrchung/S2VX/pull/461/commits/1501c22ab4b36cb2624ecd924c50b727cbb24719

Xenocidel (approved):

Naranja-Sagged (approved):

PR #473, opened 2021-05-13
Copy time to command panel

### LINKED ISSUE #470: Copy time to command pannel
As a story writer, I want to copy the current time to the command panel inputs so that I can quickly create a command at the current playback time.

- Similar to the color picker icon, add a clock icon in the StartTime and EndTime inputs

Xenocidel:
Closes #470

### DIFF
--- a/S2VX.Game/Editor/CommandPanel/CommandPanelInputBar.cs
+++ b/S2VX.Game/Editor/CommandPanel/CommandPanelInputBar.cs
@@ -11,26 +11,30 @@ using System.Collections.Generic;
 namespace S2VX.Game.Editor.CommandPanel {
     public class CommandPanelInputBar : FillFlowContainer {
         public Dropdown<string> DropType { get; } = new BasicDropdown<string> { Width = 160 };
-        public CommandPanelValueInput StartTime { get; } = new();
-        public CommandPanelValueInput EndTime { get; } = new();
+        public CommandPanelValueInput StartTime { get; }
+        public CommandPanelValueInput EndTime { get; }
         public CommandPanelValueInput StartValue { get; } = new();
         public CommandPanelValueInput EndValue { get; } = new();
         public Dropdown<string> DropEasing { get; } = new BasicDropdown<string> { Width = S2VXCommandPanel.InputSize.X };
         public Button BtnSave { get; }
 
-        public static CommandPanelInputBar CreateAddInputBar(Action<ValueChangedEvent<string>> handleTypeSelect, Action handleAddClick) =>
-            new(false, handleTypeSelect, handleAddClick);
+        public static CommandPanelInputBar CreateAddInputBar(Action<ValueChangedEvent<string>> handleTypeSelect, Action handleAddClick,
+            Func<double> currentTimeDelegate) =>
+            new(false, handleTypeSelect, handleAddClick, currentTimeDelegate);
 
-        public static CommandPanelInputBar CreateEditInputBar(Action handleSaveClick) =>
-            new(true, _ => { }, handleSaveClick);
+        public static CommandPanelInputBar CreateEditInputBar(Action handleSaveClick, Func<double> currentTimeDelegate) =>
+            new(true, _ => { }, handleSaveClick, currentTimeDelegate);
 
-        private CommandPanelInputBar(bool isEditBar, Action<ValueChangedEvent<string>> handleTypeSelect, Action handleCommitClick) {
+        private CommandPanelInputBar(bool isEditBar, Action<ValueChangedEvent<string>> handleTypeSelect, Action handleCommitClick,
+            Func<double> currentTimeDelegate) {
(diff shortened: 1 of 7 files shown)


Naranja-Sagged (approved):

S2VX.Game/Editor/CommandPanel/CommandPanelValueInput.cs
```diff
@@ -61,7 +87,9 @@ public class CommandPanelValueInput : Container {
 
         // Bindings need to be set up early so that they can trigger before
         // Load() happens
-        public CommandPanelValueInput() {
+        public CommandPanelValueInput(Func<double> currentTimeDelegate = null) {
+            CurrentTimeDelegate = currentTimeDelegate;
+            BtnApplyCurrentTime.Action = ApplyCurrentTime;
```
maxrchung, line 92: If you want, I don't think you need to store `currentTimeDelegate` as a property. You can update `ApplyCurrentTime` to take the delegate.

```suggestion
            BtnApplyCurrentTime.Action = () => ApplyCurrentTime(currentTimeDelegate);
```

S2VX.Game/Editor/CommandPanel/CommandPanelValueInput.cs
```diff
@@ -23,6 +43,10 @@ public class CommandPanelValueInput : Container {
             Position = new(0, S2VXCommandPanel.InputSize.Y)
         };
 
+        public Func<double> CurrentTimeDelegate { get; private set; }
+
+        public void UseApplyCurrentTime() => BtnApplyCurrentTime.Show();
```
maxrchung, line 48: I wouldn't explicitly call `UseApplyCurrentTime` to show the time button. I can see that it's following the way `UseColorPicker` is written, but `UseColorPicker` was written that particular way it was because when you switch to a different type of command, you have to decide at that moment whether to show/hide the color picker. The time button is different because you know at construction time whether to show or hide this button. I would put the show/hide time button logic into the constructor. You can just check if the passed in delegate is null or not.

S2VX.Game/Editor/CommandPanel/CommandPanelInputBar.cs
```diff
@@ -120,12 +114,7 @@ public class CommandPanelInputBar : FillFlowContainer {
                 }
             });
 
-        private void AddTabbableInput(string text, TabbableContainer input) {
-            AddInput(text, input);
-            input.TabbableContentContainer = this;
-        }
-
-        private void AddValueInput(string text, CommandPanelValueInput input) {
+        private void AddTextInput(string text, CommandPanelValueInput input) {
```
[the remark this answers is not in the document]
Xenocidel: Implemented

S2VX.Game/Editor/CommandPanel/CommandPanelValueInput.cs
Xenocidel: Implemented

S2VX.Game/Editor/CommandPanel/CommandPanelValueInput.cs
Xenocidel: Implemented

S2VX.Game/Editor/CommandPanel/CommandPanelInputBar.cs
```diff
@@ -4,41 +4,37 @@
...
-        public static CommandPanelInputBar CreateAddInputBar(Action<ValueChangedEvent<string>> handleTypeSelect, Action handleAddClick) =>
-            new(false, handleTypeSelect, handleAddClick);
+        public static CommandPanelInputBar CreateAddInputBar(Action<ValueChangedEvent<string>> handleTypeSelect, Action handleAddClick,
+            Func<double> currentTimeDelegate) =>
+            new(false, handleTypeSelect, handleAddClick, currentTimeDelegate);
 
-        public static CommandPanelInputBar CreateEditInputBar(Action handleSaveClick) =>
-            new(true, _ => { }, handleSaveClick);
+        public static CommandPanelInputBar CreateEditInputBar(Action handleSaveClick, Func<double> currentTimeDelegate) =>
+            new(true, _ => { }, handleSaveClick, currentTimeDelegate);
```
maxrchung, line 26: Just noticed this, but it looks like `handleTypeSelect` does nothing right now for edit input bar. This might be pretty weird for someone trying to use this drop down in the edit input bar. Not sure if we need to really address this though.

maxrchung (approved):

S2VX.Game/Editor/CommandPanel/CommandPanelInputBar.cs
Xenocidel: That's right, `HandleTypeSelect` will only be called from the `AddInputBar` because it needs to reload the command list to apply the new filter. No such action needs to be done when changing the type for the edit bar.
image

PR #504, opened 2021-06-09
Keeping track of statistics

### LINKED ISSUE #501: Keeping track of statistics
As a player, I want to keep track of statistics to show how well I did playing a map.

- [x] Total score
- [x] Accuracy: %perfect out of all notes
- [x] Number of early hits
- [x] Number of late hits
- [x] Number of perfect range hits
- [x] Number of misses
- [x] Combo
- [x] Median hit overall

maxrchung:
Closes #501

### DIFF
--- a/S2VX.Game.Tests/HeadlessTests/ScoreProcessorTests/ProcessHitCursorTests.cs
+++ b/S2VX.Game.Tests/HeadlessTests/ScoreProcessorTests/ProcessHitCursorTests.cs
@@ -1,13 +1,9 @@
 ﻿using NUnit.Framework;
 using osu.Framework.Allocation;
-using osu.Framework.Audio;
-using osu.Framework.Screens;
 using osu.Framework.Testing;
-using S2VX.Game.Play;
 using S2VX.Game.Play.UserInterface;
 using S2VX.Game.Story;
 using S2VX.Game.Story.Note;
-using System.IO;
 
 namespace S2VX.Game.Tests.HeadlessTests.ScoreProcessorTests {
     [HeadlessTest]
@@ -15,26 +11,24 @@ namespace S2VX.Game.Tests.HeadlessTests.ScoreProcessorTests {
         [Resolved]
         private S2VXCursor Cursor { get; set; }
 
-        private PlayScreen PlayScreen { get; set; }
+        [Cached]
+        private S2VXStory Story { get; } = new();
+
         private Notes Notes { get; set; }
-        private ScoreProcessor GetProcessor() => PlayScreen.ScoreProcessor;
+        private ScoreProcessor Processor { get; } = new();
 
         [BackgroundDependencyLoader]
-        private void Load(AudioManager audio) {
-            var story = new S2VXStory();
-            var audioPath = Path.Combine("TestTracks", "10-seconds-of-silence.mp3");
-            Add(new ScreenStack(PlayScreen = new PlayScreen(false, story, S2VXTrack.Open(audioPath, audio))));
-            Notes = story.Notes;
-        }
+        private void Load() => Add(Processor);
 
         [SetUpSteps]
         public void SetUpSteps() {
+            Notes = Story.Notes;
             AddStep("Reset cursor", () => Cursor.Reset());
-            AddStep("Reset score processor", () => GetProcessor().Reset());
+            AddStep("Reset score processor", () => Processor.Reset());
         }
 
         private void ProcessHit(double scoreTime) =>
-            AddStep("Process note", () => GetProcessor().ProcessHit(scoreTime, 0));
+            AddStep("Process note", () => Processor.ProcessHit(scoreTime, 0));
 
         [Test]
         public void ProcessHit_PerfectHit_ColorsCursorPerfect() {
--- a/S2VX.Game.Tests/HeadlessTests/ScoreProcessorTests/ScoreStatisticsTests.cs
+++ b/S2VX.Game.Tests/HeadlessTests/ScoreProcessorTests/ScoreStatisticsTests.cs
@@ -10,6 +10,7 @@ namespace S2VX.Game.Tests.HeadlessTests.ScoreProcessorTests {
     public class ScoreStatisticsTests : S2VXTestScene {
         [Cached]
         private S2VXStory Story { get; } = new();
+
         private Notes Notes { get; set; }
         private ScoreProcessor Processor { get; } = new();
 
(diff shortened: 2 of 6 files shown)


S2VX.Game.Tests/HeadlessTests/ScoreProcessorTests/ScoreStatisticsTests.cs
```diff
@@ -0,0 +1,140 @@
...
+        private ScoreProcessor Processor { get; } = new();
+
+        [BackgroundDependencyLoader]
+        private void Load() => Add(Processor);
+
+        [SetUpSteps]
+        public void SetUpSteps() {
+            // Tried to put this line into Load() but ran into an issue where if
+            // I ran a test by itself through the Test Explorer, the Load()
+            // wasn't triggered correctly
```
Xenocidel, line 23: I'm still getting an error when running a single test
image

Xenocidel (changes requested):

S2VX.Game.Tests/HeadlessTests/ScoreProcessorTests/ScoreStatisticsTests.cs
maxrchung: That looks like it's for other ScoreProcessor tests. I'll go and fix the other tests.

maxrchung: Refactored other files

Xenocidel (approved):

S2VX.Game.Tests/HeadlessTests/ScoreProcessorTests/ProcessHitCursorTests.cs
```diff
@@ -1,40 +1,34 @@
...
-using System.IO;
 
 namespace S2VX.Game.Tests.HeadlessTests.ScoreProcessorTests {
     [HeadlessTest]
     public class ProcessHitCursorTests : S2VXTestScene {
         [Resolved]
         private S2VXCursor Cursor { get; set; }
 
-        private PlayScreen PlayScreen { get; set; }
+        [Cached]
```
Naranja-Sagged, line 14: I might've forgot, but why do we need cached here? I don't see where this would be resolved. Does it have something to do with dependencies?

S2VX.Game.Tests/HeadlessTests/ScoreProcessorTests/ScoreStatisticsTests.cs
```diff
@@ -0,0 +1,141 @@
...
+using S2VX.Game.Story;
+using S2VX.Game.Story.Note;
+
+namespace S2VX.Game.Tests.HeadlessTests.ScoreProcessorTests {
+    [HeadlessTest]
+    public class ScoreStatisticsTests : S2VXTestScene {
+        [Cached]
+        private S2VXStory Story { get; } = new();
+
+        private Notes Notes { get; set; }
```
Naranja-Sagged, line 14: Is it possible to initialize this as = Story.Notes here so you don't have to do it in the set up steps? Might be applicable to other files in this change if this works.

S2VX.Game/Play/UserInterface/ScoreProcessor.cs
```diff
@@ -27,6 +28,30 @@ public class ScoreProcessor : CompositeDrawable {
...
 
+        public int PerfectCount { get; private set; }
+        public int EarlyCount { get; private set; }
+        public int LateCount { get; private set; }
+        public int MissCount { get; private set; }
+        public SortedList<double> Scores { get; private set; } = new SortedList<double>();
+        public int Combo { get; private set; }
+        public int MaxCombo { get; private set; }
+
+        public double Accuracy() => (double)PerfectCount / Scores.Count;
```
Naranja-Sagged, line 39: May be more appropriate to use an Expression-Bodied Member(https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/statements-expressions-operators/expression-bodied-members) here by removing the parentheses since there's no setting being done, so a property with only a getter would make sense.

S2VX.Game/Play/UserInterface/ScoreProcessor.cs
```diff
@@ -27,6 +28,30 @@ public class ScoreProcessor : CompositeDrawable {
         public S2VXSample Hit { get; private set; }
         public S2VXSample Miss { get; private set; }
 
+        public int PerfectCount { get; private set; }
+        public int EarlyCount { get; private set; }
+        public int LateCount { get; private set; }
+        public int MissCount { get; private set; }
+        public SortedList<double> Scores { get; private set; } = new SortedList<double>();
```
Naranja-Sagged, line 35: Only gonna comment this on newly added code, but is new() possible here?

S2VX.Game.Tests/HeadlessTests/ScoreProcessorTests/ScoreStatisticsTests.cs
maxrchung: No, this is not possible. You can't set properties based on other properties because the order of properties in a class is not guaranteed. Think of this possible scenario that wouldn't make sense:
```csharp
class Example {
    string a = new string(b);
    string b = new string(a);
}
```

S2VX.Game.Tests/HeadlessTests/ScoreProcessorTests/ProcessHitCursorTests.cs
maxrchung: Tip for reviewing. Select a group of lines or start your review point lower because from the preview window, I can't see the variable you're talking about.

In this case, `Story` needs to be resolved by `ScoreProcessor`.

S2VX.Game/Play/UserInterface/ScoreProcessor.cs
maxrchung: Good to know. I like the shorter syntax, so I changed to using it. However, it kind of messes with my general conception of a property. I tend of think of properties as just direct values that you return, versus functions (with the parentheses), as something that calculates a result.

S2VX.Game/Play/UserInterface/ScoreProcessor.cs
maxrchung: Changed

S2VX.Game.Tests/HeadlessTests/ScoreProcessorTests/ProcessHitCursorTests.cs
Naranja-Sagged: Oh ok. That makes sense, thanks.

S2VX.Game/Play/UserInterface/ScoreProcessor.cs
Naranja-Sagged: Pog

S2VX.Game.Tests/HeadlessTests/ScoreProcessorTests/ScoreStatisticsTests.cs
Naranja-Sagged: Oh aite

Naranja-Sagged (approved):

PR #506, opened 2021-06-19
Leaderboard in Song Preview

### LINKED ISSUE #491: Local leaderboard in song preview
As a player, I want to have a local leaderboard so I can record how well I am doing and see my past progress.

* Put scroll container in middle of Song Preview screen
* Put also in end screen
* We want to maybe think about adding some kind of textbox (that's saved as configuration?) to enter name so that we can have a kind of arcade style local leaderboard saved

Xenocidel:
Closes #491


S2VX.Game/Leaderboard/LeaderboardContainer.cs
```diff
@@ -0,0 +1,68 @@
...
+using osu.Framework.Graphics.Sprites;
+using osuTK.Graphics;
+using System.IO;
+
+namespace S2VX.Game.Leaderboard {
+    public class LeaderboardContainer : BasicScrollContainer {
+
+        private const string DefaultLeaderboardFileName = "leaderboard.json";
+        private TextFlowContainer NameColumn { get; set; }
+        private TextFlowContainer ScoreColumn { get; set; }
```
maxrchung, line 13: Can remove `set;` if these are only getting set in the constructor

S2VX.Game/Leaderboard/LeaderboardContainer.cs
```diff
@@ -0,0 +1,68 @@
...
+            var leaderboardPath = Path.Combine(Path.GetDirectoryName(storyPath), leaderboardFileName);
+            if (File.Exists(leaderboardPath)) {
+                LoadLeaderboard(leaderboardPath);
+            }
+        }
+
+
+        private void LoadLeaderboard(string leaderboardPath) {
+            var text = File.ReadAllText(leaderboardPath);
+            var data = JsonConvert.DeserializeObject<LeaderboardEntries>(text);
```
maxrchung, line 55: I think it would make more sense to do `DeserializeObject<IEnumerable<LeaderboardEntry>>(text)` here instead of having a separate `LeaderboardEntries` class.

S2VX.Game/Leaderboard/LeaderboardContainer.cs
```diff
@@ -0,0 +1,68 @@
...
+            try {
+                foreach (var entry in data.Entries) {
+                    NameColumn.AddParagraph(entry.Name);
+                    ScoreColumn.AddParagraph(entry.Score);
+                    ++EntryCount;
+                }
+            } catch (System.NullReferenceException) {
+                NameColumn.AddParagraph("Malformed or corrupted Leaderboard file!");
+                EntryCount = -1;
```
Naranja-Sagged, line 64: Might be useful to use [nullable value types syntax](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/nullable-value-types) here. I just found this from a C# tutorial yesterday but I think it allows values to be nullable so it might be possible to avoid using try each.

Xenocidel: I've changed this to catch all exceptions since it may not just be null. The deserialization could fail due to invalid or unexpected json format. But if it were just null every time then yeah the null coalescing operator would be a good one to use.

S2VX.Game/Leaderboard/LeaderboardContainer.cs
Xenocidel: Implemented

S2VX.Game/Leaderboard/LeaderboardContainer.cs
Xenocidel: Implemented

maxrchung (approved):

S2VX.Game/Leaderboard/LeaderboardContainer.cs
```diff
@@ -0,0 +1,70 @@
...
+                var data = JsonConvert.DeserializeObject<IEnumerable<LeaderboardEntry>>(text);
+                foreach (var entry in data) {
+                    NameColumn.AddParagraph(entry.Name);
+                    ScoreColumn.AddParagraph(entry.Score);
+                    ++EntryCount;
+                }
+            } catch (Exception) {
+                NameColumn.AddParagraph("Malformed or corrupted Leaderboard file!");
+                EntryCount = -1;
+            }
```
maxrchung, line 67: To follow the other `Exception` handling examples, I'd like to console print out the `Exception.ToString()` so we know at least in the logs what exact Exception happened.

S2VX.Game/Leaderboard/LeaderboardContainer.cs
maxrchung: I'm not a fan of nullable types. Feels like it just introduces a bigger can of worms to handle more cases of null.

S2VX.Game/Leaderboard/LeaderboardContainer.cs
Xenocidel: Implemented

maxrchung (approved):

Naranja-Sagged (approved):

S2VX.Game.Tests/UnitTests/LeaderboardTests/LoadLeaderboardTests.cs
```diff
@@ -0,0 +1,32 @@
...
+        }
+
+        [Test]
+        public void ValidStoryboardFile_HasThreeEntries() {
+            var input = "valid.json";
+            var leaderboard = new LeaderboardContainer(StoryPath, leaderboardFileName: input);
+            Assert.AreEqual(3, leaderboard.EntryCount);
+        }
+    }
+}
```
maxrchung, line 32: Look into using headless unit test instead of unit test

Xenocidel: Changed to visual test

maxrchung (approved):

PR #511, opened 2021-07-03
Leaderboard on end game

### LINKED ISSUE #502: Local leaderboard in end game screen
As a player, I want to see and use the local leaderboard in the end game screen so that I can record my score after playing a map.

Xenocidel:
Closes #502

### DIFF
--- a/S2VX.Game/Leaderboard/AddLeaderboardEntryContainer.cs
+++ b/S2VX.Game/Leaderboard/AddLeaderboardEntryContainer.cs
@@ -1,0 +1,46 @@
+﻿using osu.Framework.Allocation;
+using osu.Framework.Graphics;
+using osu.Framework.Graphics.Containers;
+using osu.Framework.Graphics.Sprites;
+using osu.Framework.Graphics.UserInterface;
+using osuTK;
+using osuTK.Graphics;
+
+namespace S2VX.Game.Leaderboard {
+    public class AddLeaderboardEntryContainer : FillFlowContainer {
+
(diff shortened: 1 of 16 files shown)


S2VX.Game/Leaderboard/AddLeaderboardEntryContainer.cs
```diff
@@ -0,0 +1,50 @@
...
+            Children = new Drawable[] {
+                NameInput = new() {
+                    Size = new Vector2(InputWidth * 0.75f, InputHeight),
+                    BorderColour = Color4.Red,
+                    Masking = true,
+                },
+                SaveButton = new() {
+                    Action = () => {
+                        LeaderboardContainer.AddEntry(NameInput.Text, Score);
+                        Clear();
```
maxrchung, line 40: This feels kinda awkward to me. Since disabling seems somewhat difficult, how about something like this:
- Always enable save button
- Remember entry once you save
- Remove previous entry if you already saved before

Also helps with cases where someone mistyped their name and wanted to correct it.

Naranja-Sagged: Yeah this sounds like a good idea to me. The save button disappearing seemed ok to me but fixing typos would be nice.

Xenocidel: Implemented

maxrchung (approved):

Naranja-Sagged (approved):

PR #518, opened 2021-07-21
Score markers on notes

### LINKED ISSUE #427: Add score markers on notes
As a player, I want there to be score markers when I hit notes so that I can see if I am hitting notes early, late, or missing them.

Xenocidel:
Closes #427


S2VX.Game/Story/Note/Notes.cs
```diff
@@ -34,13 +34,13 @@ public class Notes : CompositeDrawable {
...
-        public float PerfectThreshold { get; set; } = 30;
-        public float HitThreshold { get; set; } = 100;
-        public float MissThreshold { get; set; } = 200;
+        public static Color4 PerfectColor { get; set; } = S2VXColorConstants.LightYellow;
+        public static Color4 EarlyColor { get; set; } = S2VXColorConstants.White;
+        public static Color4 LateColor { get; set; } = S2VXColorConstants.DarkYellow;
+        public static Color4 MissColor { get; set; } = S2VXColorConstants.Red;
+        public static float PerfectThreshold { get; set; } = 30;
+        public static float HitThreshold { get; set; } = 100;
+        public static float MissThreshold { get; set; } = 200;
```
maxrchung, line 43: In general, I think [making stateful member variables static is bad practice](https://stackoverflow.com/questions/7026507/why-are-static-variables-considered-evil), even if it cuts down on some references. We're probably okay right now with how we are carefully writing our tests and not setting these variables, but imagine this maybe hyperbolic scenario: Test1 sets MissThreshold to 999999 seconds, Test2 has nothing to do with testing miss threshold but now behaves/works incorrectly due to Test1's configuration.

At the very least, I would remove all the setters and treat these as constants.

S2VX.Game/Story/Note/HitMarker.cs
```diff
@@ -0,0 +1,57 @@
...
+        public bool UpdateMarker() {
+            if (Alpha == 0) {
+                return true;
+            }
+            UpdatePosition();
+            UpdateAlpha();
+            return false;
+        }
+
+        private void UpdateAlpha() => Alpha = S2VXUtils.ClampedInterpolation(Time.Current, MarkerAlpha, 0.0f, SpawnTime, SpawnTime + 200);
```
maxrchung, line 41: Maybe we want to pull this out to Notes or somewhere so we can more easily keep track of Note related configurations.

S2VX.Game/Story/Note/HitMarkers.cs
```diff
@@ -0,0 +1,49 @@
+﻿using osu.Framework.Allocation;
+using osu.Framework.Graphics;
+using osu.Framework.Graphics.Containers;
+using osuTK;
+using osuTK.Graphics;
+using System.Collections.Generic;
+
+namespace S2VX.Game.Story.Note {
+    public class HitMarkers : CompositeDrawable {
```
maxrchung, line 9: I have some general comments after some playtesting:
* Miss markers don't show up
* Let hit markers stay longer, osu! has a pretty long window for this I think
* Change hit marker colors to be more contrasting, e.g. blue, orange maybe
* Remove hit marker for perfect
* Remove color changes on cursor

Naranja-Sagged: I agree with all of these points. Right now it's pretty hard to tell if I'm late or early just by the color since perfect is yellow, and  slightly late&early are different shades of yellow. Having big contrast between late & early + the current shade changing would be helpful

S2VX.Game/Story/Note/Notes.cs
Xenocidel: Ah yeah, I was thinking these would never be changed. Missed removing the setters.

S2VX.Game/Story/Note/HitMarker.cs
Xenocidel: You mean move MarkerAlpha and 200 (let's call it MarkerShowTime) to a property of Notes?

S2VX.Game/Story/Note/HitMarkers.cs
Xenocidel: - Miss markers seem to be working for me
- I'll change to 800ms and see how it is
- I've changed early to Cyan and late to Magenta
- Removed perfect
- Removed cursor color changes

S2VX.Game/Story/Note/HitMarker.cs
maxrchung: Yeah

maxrchung (approved):

Naranja-Sagged (approved):

PR #519, opened 2021-07-26
Multi point hold notes

### LINKED ISSUE #486: Multi-point hold notes
As a player I want hold notes to be multi-pointed so that I can have more interesting gameplay

maxrchung:
Closes #486


maxrchung:
Note that the multi-point hold note feature has been split up into separate stories. This PR is for checking in the multi-point implementation that follows how osu! implements its sliders. This change should be completely backwards compatible with our existing maps and behavior.

Issue #520 will implement the editor side changes that allows you to actually edit multi-points on hold notes.

S2VX.Game/Story/Note/HoldNote.cs
```diff
@@ -22,20 +23,64 @@ public abstract class HoldNote : S2VXNote {
...
         protected override void UpdatePosition() {
             UpdateIndicator();
             UpdateSliderPath();
         }
 
+        // These are static helpers for calculating hold related values
+        // TODO: There's probably a better way to share this between HoldApproach and HoldNote for calculating these values
+        // TODO: Optimize these so that they are called once on initialization instead of ran every update loop
+        private static IEnumerable<Vector2> CombineAllCoordinates(Vector2 coordinates, List<Vector2> midCoordinates, Vector2 endCoordinates)
+            => new List<Vector2> { coordinates }.Concat(midCoordinates.Append(endCoordinates));
```
Xenocidel, line 44: this is just a comment. lmao

S2VX.Game/Story/Note/SliderPathDrawNode.cs
```diff
@@ -0,0 +1,246 @@
...
+        private class SliderPathDrawNode : DrawNode {
+            protected new SliderPath Source => (SliderPath)base.Source;
+
+            private readonly List<Line> Segments = new();
+
+            private Vector2 DrawSize;
+            private float Radius;
+            private IShader PathShader;
+
+            private readonly LinearBatch<TexturedVertex3D> LinearBatch = new(4 * 100 * 3, 10, PrimitiveType.Triangles);
```
Xenocidel, line 29: Are these numbers supposed to mean something? Can we add a comment explaining them

S2VX.Game/Story/Note/SliderPathDrawNode.cs
```diff
@@ -0,0 +1,246 @@
...
+            }
+
+            private void AddLineQuads(Line line, RectangleF texRect) {
+                var closestCorner = FindClosestCorner(line.OrthogonalDirection);
+                var oppositePoint = new Vector2(-closestCorner.X, -closestCorner.Y);
+                var lineLeft = new Line(line.StartPoint + closestCorner, line.EndPoint + closestCorner);
+                var lineRight = new Line(line.StartPoint + oppositePoint, line.EndPoint + oppositePoint);
+
+                var screenLineLeft = new Line(Vector2Extensions.Transform(
+                    lineLeft.StartPoint, DrawInfo.Matrix),
```
Xenocidel, line 97:
```suggestion
                var screenLineLeft = new Line(
                    Vector2Extensions.Transform(lineLeft.StartPoint, DrawInfo.Matrix),
```

Xenocidel (changes requested):

S2VX.Game/Story/Note/SliderPathDrawNode.cs
maxrchung: Fixed

S2VX.Game/Story/Note/SliderPathDrawNode.cs
maxrchung: I don't really know what these do. I put in osu!'s comment that they had in Path_DrawNode.cs:

```csharp
            // We multiply the size param by 3 such that the amount of vertices is a multiple of the amount of vertices
            // per primitive (triangles in this case). Otherwise overflowing the batch will result in wrong
            // grouping of vertices into primitives.
            private readonly LinearBatch<TexturedVertex3D> halfCircleBatch = new LinearBatch<TexturedVertex3D>(MAX_RES * 100 * 3, 10, PrimitiveType.Triangles);
            private readonly QuadBatch<TexturedVertex3D> quadBatch = new QuadBatch<TexturedVertex3D>(200, 10);
```

S2VX.Game/Story/Note/HoldNote.cs
```diff
@@ -22,20 +23,64 @@ public abstract class HoldNote : S2VXNote {
...
+            }
+            return totalDistance;
+        }
+        public static Vector2 InterpolateCoordinates(
+            double currentTime,
+            double hitTime,
+            double endTime,
+            Vector2 startCoordinates,
+            List<Vector2> midCoordinates,
+            Vector2 endCoordinates
```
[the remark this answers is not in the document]
maxrchung: In-line is my preference, but for this function, adding arguments in-line will cross the 160 character limit.

S2VX.Game/Story/Note/HoldNote.cs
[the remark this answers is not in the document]
maxrchung: The return type of `Append()` is `IEnumerable`. When working with LINQ functions, it can be easier to treat everything as `IEnumerable` since LINQ expects to work with `IEnumerable` as inputs and outputs. Alternatively, you could call `ToList()` at the end and convert it to a list:

```csharp
private static List<Vector2> CombineAllCoordinates(Vector2 coordinates, List<Vector2> midCoordinates, Vector2 endCoordinates)
    => new List<Vector2> { coordinates }.Concat(midCoordinates.Append(endCoordinates)).ToList();
```

But I don't really like this since it feels like it's wasting a step to do nothing.

maxrchung: I left these in intentionally. For this change, I didn't want to spend too much time refactoring the structure of Notes and Approaches. These are suggestions to optimize some performance. We can make stories from these, but I wouldn't want to put them into any upcoming milestones since they're not really needed.

Xenocidel: nah was just chuckling at the way this was constructed

S2VX.Game/Story/Note/HoldNote.cs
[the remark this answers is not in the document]
Xenocidel: Oh ok, I thought we would just go to a second line then, rather than having one line for each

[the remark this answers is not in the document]
maxrchung: I think it's clearer to me if the arguments are either all in one line or each in a line. Not a fan of putting multiple arguments in a single line since I think it can be easy to missread.

Xenocidel (approved):

Naranja-Sagged (approved):

PR #521, opened 2021-08-01
Global volume control

### LINKED ISSUE #155: Global volume control
As a player I want to be able to change the volume of the game outside of just the editor so I don't have to go into the editor to adjust the volume.

Naranja-Sagged:
Closes #155


S2VX.Game.Tests/HeadlessTests/GameHoldNoteTests.cs
```diff
@@ -15,6 +15,8 @@
 namespace S2VX.Game.Tests.HeadlessTests {
     [HeadlessTest]
     public class GameHoldNoteTests : S2VXTestScene {
+        [Cached]
+        private GlobalVolumeDisplay VolumeDisplay { get; set; } = new();
```
maxrchung, line 19: Does this need to be cached within every single test? I think you can save a lot of redundancy if you cache it once somewhere, maybe in `S2VXTestScene`.

S2VX.Game/S2VXScrollContainer.cs
```diff
@@ -0,0 +1,10 @@
+﻿using osu.Framework.Graphics.Containers;
+using osu.Framework.Input.Events;
+
+namespace S2VX.Game {
+    public class S2VXScrollContainer : BasicScrollContainer {
+        public S2VXScrollContainer() { }
+
+        protected override bool OnScroll(ScrollEvent e) => false;
```
maxrchung, line 8: How is this going to work? You're basically removing all the scrolling functionality of the `BasicScrollContainer` which means scrolling won't work anymore.

Also, if you're doing it this way you need to change out all instances of `BasicScrollContainer`. I don't see `CommandPanel` in this PR, so you probably have to address that and any more instances.

As it turns out osu! uses this scroll container override hack to get by the scroll problem: https://github.com/ppy/osu/blob/7654df94f6f37b8382be7dfcb4f674e03bd35427/osu.Game/Graphics/Containers/OsuScrollContainer.cs#L101

maxrchung (changes requested):

S2VX.Game.Tests/HeadlessTests/GameHoldNoteTests.cs
Naranja-Sagged: Oh I didn't know that was possible lol. Added to S2VXTestScene.

S2VX.Game/S2VXScrollContainer.cs
Naranja-Sagged: I didn't think about that 😹. Ok, I added base.Onscroll() to it so it should be working fine now. Also switched out command panel.

Xenocidel:
If you press Esc in the editor, change the volume while in the exit confirmation screen, and cancel to go back to editor, the volume display does not update. Should we just remove that volume display and have people change volume with Alt like everywhere else?

Xenocidel:
Visual GUI GlobalVolumeDisplay tests fail if you run them multiple times
image

Naranja-Sagged:
> If you press Esc in the editor, change the volume while in the exit confirmation screen, and cancel to go back to editor, the volume display does not update. Should we just remove that volume display and have people change volume with Alt like everywhere else?

Yeah removing the editor's volume display and replacing it with the global one sounds like a good idea. @maxrchung Thoughts?

Naranja-Sagged:
> Visual GUI GlobalVolumeDisplay tests fail if you run them multiple times
> image

Could this be related to the bug I mentioned on Discord? Looks similar to the errors I was getting. The error happens whenever the song runs out of time in this case it should be 10 seconds, so it tries to bring up leaderboard which doesn't exist.

maxrchung:
> > If you press Esc in the editor, change the volume while in the exit confirmation screen, and cancel to go back to editor, the volume display does not update. Should we just remove that volume display and have people change volume with Alt like everywhere else?
> 
> Yeah removing the editor's volume display and replacing it with the global one sounds like a good idea. @maxrchung Thoughts?

I agree. I think there is also value in keeping hotkeys more consistent.

Xenocidel:
> > Visual GUI GlobalVolumeDisplay tests fail if you run them multiple times
> > image
> 
> Could this be related to the bug I mentioned on Discord? Looks similar to the errors I was getting. The error happens whenever the song runs out of time in this case it should be 10 seconds, so it tries to bring up leaderboard which doesn't exist.

I don't think so, this is the newly added tests which just scrolls to increase/decrease volume. Can you reproduce?

Naranja-Sagged:
> > > Visual GUI GlobalVolumeDisplay tests fail if you run them multiple times
> > > image
> > 
> > 
> > Could this be related to the bug I mentioned on Discord? Looks similar to the errors I was getting. The error happens whenever the song runs out of time in this case it should be 10 seconds, so it tries to bring up leaderboard which doesn't exist.
> 
> I don't think so, this is the newly added tests which just scrolls to increase/decrease volume. Can you reproduce?

image
I get this if I just leave the visual GUI sitting there on the test. This can also be reproduced in other tests, like leaving your screen on PlayInfoBarTests for example

Naranja-Sagged:
> > > If you press Esc in the editor, change the volume while in the exit confirmation screen, and cancel to go back to editor, the volume display does not update. Should we just remove that volume display and have people change volume with Alt like everywhere else?
> > 
> > 
> > Yeah removing the editor's volume display and replacing it with the global one sounds like a good idea. @maxrchung Thoughts?
> 
> I agree. I think there is also value in keeping hotkeys more consistent.

Ok, @Xenocidel @maxrchung I have removed the Editor Volume Display

S2VX.Game/Editor/Containers/EditorInfoBar.cs
```diff
@@ -42,12 +42,6 @@ public class EditorInfoBar : CompositeDrawable {
             TextAnchor = Anchor.TopCentre,
         };
 
-        public VolumeDisplay VolumeDisplay { get; } = new() {
-            Anchor = Anchor.TopRight,
-            Origin = Anchor.TopRight,
-            TextAnchor = Anchor.TopRight,
-        };
```
maxrchung, line 49: Need to also remove menu items related to volume

Also, adjusting volume with alt + mouse scroll in the command panel doesn't seem to be working for me

Naranja-Sagged: Ok menu items removed.

Lol the scroll container was only counting as the scroll bar so volume scrolling only worked there. I added onscroll to the whole command panel so it should be working now.

maxrchung: Good catch, looks good now

maxrchung (approved):

Xenocidel (approved):

PR #541, opened 2021-08-19
Tap for timing

### LINKED ISSUE #485: Tap for timing
As a story writer, I wish to be able to time a song natively on S2VX instead of borrowing offset and BPM off of osu! or timing it on osu! if it does not exist yet.

maxrchung:
Closes #485 

Did a few refactors:
* Moved CommandPanel and MetadataPanel into S2VX.Game.Editor.Containers folder
* Made a base class that CommandPanel, MetadataPanel, and TapPanel inherit from


Xenocidel:
When tapping using keypresses, the tap panel triggers even if the cursor isn't over the tap area. Is that intended?

maxrchung:
> When tapping using keypresses, the tap panel triggers even if the cursor isn't over the tap area. Is that intended?

Yeah, I noticed that. That's the default behavior of the default OnKey handlers. I kept it like that cause you'll need to do extra work to account for position. You'll probably have to do some additional special position checks if you want to account for that or use the key binding containers like the play screen.

maxrchung:
> When tapping using keypresses, the tap panel triggers even if the cursor isn't over the tap area. Is that intended?

I did some more experimentation, and I think I was able to make it more consistent with my latest changes. By using an `IsHovered` check, `OnKey` handlers now properly processes a tap only when the mouse is over the tap receptor. Also, I switched the `OnClick` handler to `OnMouseUp` because I noticed that `OnClick` actually doesn't trigger for right click for some reason.

S2VX.Game/Editor/UserInterface/TapReceptor.cs
```diff
@@ -0,0 +1,108 @@
...
+        // ProcessTap() on key up instead of key down so we lessen the chance of
+        // held presses, i.e. if you hold down the Z key, it will continuously
+        // trigger OnKeyDown
+        protected override void OnKeyUp(KeyUpEvent e) {
+            if (IsHotKeyPressed) {
+                ProcessTap();
+            }
+        }
+
+        protected override void OnMouseUp(MouseUpEvent e) {
```
Xenocidel, line 59:
```suggestion
        protected override void OnMouseDown(MouseUpEvent e) {
```
More accurate for timing since people might hold the mouse button

Xenocidel (changes requested):

S2VX.Game/Editor/UserInterface/TapReceptor.cs
maxrchung: I switched this to `OnMouseDown`. I also changed `ProcessTap()` to run in `OnKeyDown` - was able to get this to work by keeping track of held keys with a `HashSet`.

I looked into the possibility of just reusing the `PlayKeyBindingContainer` for `TapReceptor`, but that might take some weird refactoring because the key container requires a `PlayScreen` reference.

maxrchung: Also added some unit tests that specifically test key down

Xenocidel (approved):

Naranja-Sagged (approved):

PR #583, opened 2021-10-17
Fix s2vx web

### LINKED ISSUE #576: Investigate bottom left error and fix descriptions
As a website viewer, I want the site to have the proper description and no errors so that I am not confused at what's on the site.

* Bottom left is showing I'm assuming what is a Blazor error, needs to be looked at
* Descriptions should be a bit better formatted and fixed up

Naranja-Sagged:
Closes #576

### DIFF
--- a/S2VX.Web/wwwroot/css/app.css
+++ b/S2VX.Web/wwwroot/css/app.css
@@ -58,15 +58,3 @@ a, .btn-link {
     transform: translate3d(-50%, -50%, 0);
     filter: brightness(50%);
 }
-
-#blazor-error-ui {
-    background: lightyellow;
-    bottom: 0;
-    box-shadow: 0 -1px 2px rgba(0, 0, 0, 0.2);
-    display: none;
-    left: 0;
-    padding: 0.6rem 1.25rem 0.7rem 1.25rem;
-    position: fixed;
-    width: 100%;
-    z-index: 1000;
-}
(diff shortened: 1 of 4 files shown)


S2VX.Web/wwwroot/css/app.css
```diff
@@ -58,3 +58,15 @@ a, .btn-link {
...
+    background: lightyellow;
+    bottom: 0;
+    box-shadow: 0 -1px 2px rgba(0, 0, 0, 0.2);
+    display: none;
+    left: 0;
+    padding: 0.6rem 1.25rem 0.7rem 1.25rem;
+    position: fixed;
+    width: 100%;
+    z-index: 1000;
+}
```
maxrchung, line 72: This doesn't solve the problem. Scroll down and you'll still see the text there:
image

I did some quick testing, and you can remove the `blazor-error-ui` div from `S2VX.Web/wwwroot/index.html`

maxrchung (changes requested):

S2VX.Web/wwwroot/css/app.css
maxrchung: No, you should remove both the div and css sections

Naranja-Sagged: I removed both in the latest change but it still gives error 😹

